PR #4, opened 2023-11-16
Add Formula ABC

egarciamendez:
## Description

Add Equation ABC needed as basis for all Equation types on Blueprints.

Fixes # 3


blueprints/codes/__init__.py
```diff
@@ -0,0 +1,56 @@
+"""Module for the abstract base class Equation."""
+from abc import ABC, abstractmethod
+
+
+class Equation(ABC):
```
bro-wi, line 5: A nice start, but as this is a quite essential class of our library I have An alternative proposol for the solution architecture of this. --> It is a bit hard to explain in this comment, so maybe we should have a small brainstorm session with a small amount of people:

### Proposal:

It is still a a 'mindgame' for me, but look at this idea, we can take this as a start:

Let's distinguise a 'variable' class and an "equation" class. 

The variable class is essentially a float in usage with extra attributes: a variable name, and a symbol and maybe even a description (because there are many "alphas" in the EC)

The equation class is a composition of multiple variables, it actually should act like a callable (function) because I think that is the most intuitive way of using it for a user the result of an equation: a variable. The equation class itself holds a symbolic reprensentation  (if you want to use it in a report) of the equation and things like: the source document, equation label etc.

The resulting variable stores the intermediate values of the variables, the actual equation object, the source, When we design the 'variable' as a "tree" [Composite design pattern]  the 'variable' should be able to keep track of a 'equation-pipe-line' (parent-child dependecies). By using recursion you should be able to reconstruct the complete intermediate results. I'm not sure yet if it is a problem if using a variable multiple times in different equations, maybe the child parent relationsship solves this right away.

It is a bit complex, but if we can do this correctly it would be a amazing feature that can create a complete report of a calculation without the user having any trouble doing this.

But as I said: this is only something that is spinning around in my mind. Let me know what you think. Happy to discuss this to take it to a higher level :-)

blueprints/codes/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Abstract base class for equations used in the code.
+    """
+
+    @property
+    def equation_number(self) -> str:
```
bro-wi, line 11: I would suggest: equation_label or something, it is not always an actual 'number' in my opinion.

bro-wi, line 11: It seems that (by design) you didn't chain the property and abstractmethod here. My question: if the property is optionally anyway: why do we define it in the ABC? We can leave it out and use it in the actuall classes we use? Now You are poluting your documentation and intellisense suggestions with options that won't work.

bro-wi (commented):
I think A very nice start, but I have a proposol for the design of the class. I think we can really improve on the user experience, and because the object is very important for the rest of our library we have to make sure the design is right.

Also some minor comments on the abstract base class.

blueprints/codes/__init__.py
egarciamendez: Hey @bro-wi, I can imagine that it is hard to put this idea on words. I'm up for a small brainstorm somewhere this week (do you guys prefer business hours or nights?). I'm pretty much in line with you when you say we got to get this one right. :-)

blueprints/codes/__init__.py
egarciamendez: I have no particular preference on this one but last meeting most people went for equation_number

egarciamendez: You're right an equation should always have an equation_number and thus be an abstractmethod.

blueprints/codes/__init__.py
bro-wi: I'd prefer business hours, my nights are terrible at the moment (wait for it) :-p

some timeslots this week:

today: until 16.00
Wednesday: 9.00-14.00
Thursday: 9.00-14.00

egarciamendez: 11.30 today?
https://teams.microsoft.com/l/meetup-join/19%3ameeting_NzEzNGI0YTEtOGM0Yy00ZTgyLWFmOTAtYzgyZTdhNDA2ZDcy%40thread.v2/0?context=%7b%22Tid%22%3a%22bf5f4046-a1dc-4119-aa8a-bfb9fbf46271%22%2c%22Oid%22%3a%22b1980a8f-5dff-4ec7-b9f5-d95ea2af341d%22%7d

bro-wi: As discussed, I tried prototyping some stuff, 
 this is what I came up with:

# First of all:
IDEA: can we include the design decisions in the Issues? for example: why do we choose to use "Equation" as a name and not "Formula", the solution architecture and interface we choose to use etc. etc. 

I think it is good to have a discussions on those matters prior to coding, now we are discussing in a piece of code that is already done: it makes it hard to move directions.



# Proposals:
I know these proposals are incomplete and absolutely not on a polish level, but I quickly wanted to test de functionality without worrying about variable_names, structures etc --> this can be done when we choose the actual implementation.

## proposal 1: close to the curret implementation
In this proposal I kept it as close to the current impelemtation  with the extension that you are able to use a formula as input value.

Benefits:
- return types of methods are quite flexible to implement for each function
- design is quite simple

Downsides:
- using the result is quite counter intuitive (it is not a float or something: you must use class.result to use the actual result
- I don't like te use a class that is actually a function (But that is a matter of taste)

``` 
# ? Proposal 1, closest to current design
# TODO: Design choice: is it an Equation, Expression, Formula , Identity --> https://www.mathswithgraham.org.uk/?p=2873
class Formula:
    equation_label = "6.1"
    source_document = "EC15"

    def __init__(self, *, a: float | Self, b: float | Self, c: float | Self) -> None:
        self.a = a.result() if isinstance(a, Formula) else a
        self.b = b.result() if isinstance(b, Formula) else b
        self.c = c.result() if isinstance(c, Formula) else c
        self.vars = {"a": a, "b": b, "c": c}

    def result(self) -> float:
        return self.a + self.b + self.c

    def detailed_result(self) -> dict[str, Any]:
        calculation_steps = {
            "formula": f"X = a + b + c = {self.result()}    ({self.equation_label})"
        }

        for var, value in self.vars.items():
            if isinstance(value, Formula):
                calculation_steps[var] = value.detailed_result()
            else:
                calculation_steps[var] = value

        # optionally extend these results with specific intermediate results

        return calculation_steps


# when using this structure you can retrieve the complete "function pipeline"
x = Formula(a=1, b=3, c=3)
y = Formula(a=1, b=x, c=1)
y.equation_label = "6.2"  # changing the label for better traceable results
z = Formula(a=y, b=x, c=4)
z.equation_label = "6.3"

for formula in (x, y, z):
    print(f"detailed answer for {formula.equation_label}")
    print(json.dumps(formula.detailed_result(), indent=4))
    print(10 * "=")
```

## proposal 2: the function object acts like a float
In this proposal the function is acting like a float and the result can thus be used directly (i.o using class.result)

Benefits:
- design is quite simple
- usage of the result is intuitive and easy

Downsides:
- I don't like te use a class that is actually a function (But that is a matter of taste), but in this case it feels more like a function because you don't need the ".result" notation.
- when multiple return types are possible we have to create multiple objects (not sure if this actually happens)

``` 
# ? Proposal 2, Class where the formula is acting like a float so it can be used directly without using the .result() method.
class FormulaAsFloat(float):
    equation_label = "6.1"
    source_document = "EC15"

    def __new__(cls, *, a, b, c) -> Self:
        return float.__new__(cls, a + b + c)

    def __init__(self, *, a: float | Self, b: float | Self, c: float | Self) -> None:
        self.a = a
        self.b = b
        self.c = c
        self.vars = {"a": a, "b": b, "c": c}

    def detailed_result(self) -> dict[str, Any]:
        calculation_steps = {
            "formula": f"X = a + b + c = {self}    ({self.equation_label})"
        }

        for var, value in self.vars.items():
            if isinstance(value, FormulaAsFloat):
                calculation_steps[var] = value.detailed_result()
            else:
                calculation_steps[var] = value

        # Optionally extend these results with specific intermediate results

        return calculation_steps


# when using this structure you can retrieve the complete "function pipeline"
x = FormulaAsFloat(a=1, b=3, c=3)
y = FormulaAsFloat(a=1, b=x, c=1)
y.equation_label = "6.2"  # changing the label for better traceable results
z = FormulaAsFloat(a=y, b=x, c=4)
z.equation_label = "6.3"

for formula in (x, y, z):
    print(f"detailed answer for {formula.equation_label}")
    print(f"answer={formula}") #here you see that it acts like a float, we could even implement a __str__ method with units for the answer
    print(json.dumps(formula.detailed_result(), indent=4))
    print(10 * "=")
```


## proposal 3: using a function, but with function attributes

This one is not worked out properly yet, and i'm not really sure if this is what we want: but maybe it inspires someone else.
This could be an idea: It could work if we just want to create a function that outputs something and the function has function attributes with the equation_label and document--> on the "function-pipeline" part this takes a bit more work but generally the idea could be something like this:


``` 
class FunctionOutput(float):
    def __new__(cls, value, **kwargs) -> Self:
        return float.__new__(cls, value)

    def __init__(self, value, *, formula, detailed_results):
        self.formula = formula
        self.detailed_results = detailed_results


def formula_test(*, a, b, c):
    formula_test.label = "6.10"
    formula_test.source_document = "EC512540"

    return FunctionOutput(a + b + c, formula=formula_test, detailed_results="this")


result = formula_test(a=1, b=2, c=3)
print(result)
print(result.formula.label)
```

Just let me know what everybody else think, if these designs are undesirable we just ditch it, but In that case at least we have investigated multiple options.

bro-wi, line 5: Why dit we choose "Equation"? it doesn't seem corresponding to the definitions in math: https://www.mathswithgraham.org.uk/?p=2873

I'm happy to choose whatever name we want: but I think it is a good idea to document the design choice in the Issue #4  so we can look it up later and avoid later discussions.

egarciamendez: @bro-wi sharp!! In this case it should be Formula instead of Equation. I will change it 👌

egarciamendez: @bro-wi, great work on developing these proposals! You truly think outside the box :-)

I like the solutions presented in the first two proposals. I tried to improve upon them by adding a clear interface to all subclasses of `Formula`, using an Abstract Base Class (ABC). In this new design, the result method is specifically used to separate the logic when using the __new__ method, which helps in maintaining a clear structure and readability. What do you guys think @Blueprints-org/maintainers? I believe we're almost there.

Pros:
- Consistency: Ensures a standard structure for all formula subclasses.
- Separation of Logic: Distinctly separates calculation logic (result) from object instantiation (__new__).
- Flexible Implementation: Allows different calculation logic for various formulas.
- Intuitive Usage: Operates like a float, making it user-friendly. No need to use .result().
- Detailed Documentation: Supports comprehensive representation and documentation of formulas.

Cons:

- Complex for Simple Tasks: May be overly complex for basic calculations.
- Learning Curve: Could require additional time for new users to learn.
- Limited to Float Results: Primarily designed for float outputs, might need adjustments for other return types.


```python
from abc import ABC, abstractmethod
from typing import Any


class Formula(float, ABC):
    def __new__(cls, *args, **kwargs) -> 'Formula':
        result = cls.result(*args, **kwargs)
        return float.__new__(cls, result)

    @staticmethod
    @abstractmethod
    def result(*args, **kwargs) -> float:
        pass

    @abstractmethod
    def view_latex(self) -> str:
        pass

    @abstractmethod
    def source_document(self) -> str:
        pass

    @abstractmethod
    def equation_label(self) -> str:
        pass


class Form5Dot2SpecificFormula(Formula):
    equation_label = "5.2"
    source_document = "EC2 1992-1-1"

    def __init__(self, a: float, b: float, additional_param: float) -> None:
        self.a = a
        self.b = b
        self.additional_param = additional_param
        self.vars = {"a": a, "b": b, "additional_param": additional_param}

    @staticmethod
    def result(a: float, b: float, additional_param: float) -> float:
        """Main logica calculation to be performed."""
        return a + b + additional_param

    def view_latex(self) -> str:
        """One of the possible available methods to represent the formula."""
        return f"X = {self.a} + {self.b} + {self.additional_param} = {self}"


class Form7Dot8AnotherFormula(Formula):
    equation_label = "7.8"
    source_document = "EC2 1992-1-1"

    def __init__(self, q: float, p: float, r: float) -> None:
        self.q = q
        self.p = p
        self.r = r
        self.vars = {"q": q, "p": p, "r": r}

    @staticmethod
    def result(q: float, p: float, r: float) -> float:
        """Main logica calculation to be performed."""
        return (q + p) / r

    def view_latex(self) -> str:
        """One of the possible available methods to represent the formula."""
        return f"X = ({self.q} + {self.p}) / {self.r} = {self}"


# Example usage
form_5_2 = Form5Dot2SpecificFormula(a=1, b=3, additional_param=2)
form_7_8 = Form7Dot8AnotherFormula(q=1, p=form_5_2, r=2)
print(f"Result of form_5_2: {form_5_2}")  # Output will be the float result of the calculation
print(f"Result of form_7_8: {form_7_8}")  # Output will be the float result of the calculation
print(f"LaTeX representation of form_5_2: {form_5_2.view_latex()}")
print(f"LaTeX representation of form_7_8: {form_7_8.view_latex()}")

# Result of form_5_2: 6.0
# Result of form_7_8: 3.5
# LaTeX representation of form_5_2: X = 1 + 3 + 2 = 6.0
# LaTeX representation of form_7_8: X = (1 + 6.0) / 2 = 3.5

```

P.S>I'm aware I'm missing the chaining of detailed_results which I liked on your second proposal. I'm so sad a day has only 24 hours :-(

bro-wi: Haha, don't be sad!
I like your additions, and totally agree about separating initiation vs calculation.

Two suggestions:
1. I would suggest to change the ".result()" method into ".evaluate()" as your are performing an calculation. It feels more like an 'action': but it is a matter of taste.
2. we could consider to introduce a "Variable" class (also a float) with a "view_ latex" method (maybe even a unit?)  --> Then you can use that directly in the 'view_latex' method in the formula. That is possibly useful when multiple variables are used in multiple formula's (e.g. rho etc.). Maybe that is overkill for now...


I also agree on the cons, it is a bit complex for developing, (although when the ABC is already there, we already have taken care of that aspect). So we have to weigh if the pros against the cons.

About the the 'only-float' results. It is true, for some types it is doable (when we change the order of succesion [so the __new__ a the subclass) but:
if you would like to create a bool type it is not even possible: the boolean is not subclassable because it is a Singleton. https://mail.python.org/pipermail/python-dev/2002-March/020822.html

I cant really overview how often another return type (other than float) would occur... who can jump in on that?

egarciamendez: I agree, evaluate() instead of result() maybe makes more sense. Glad we're coming to a solution. I will work out the details and push them to this PR again. Hopefully we can start soon writing the Formulas :-)

egarciamendez: Done ;-)

johan-tuls:
Also using a @staticmethod for the evaluate method prevents ever using "self". So I wonder how you would like to apply certain intermediate values without duplicating logic. 
A property method isn't possible, so I assume all should be define inside the init as vars?

SZeltaat (commented):
Great work guys!
I really like the new structure. I guess every implementation has some pro's and con's. My only concern with this implementation is that it might prevent using units. See my comment on that.

blueprints/codes/formula.py
```diff
@@ -0,0 +1,62 @@
...
+
+class Formula(float, ABC):
+    """
+    Abstract base class for formulas used in the codes.
+    """
+
+    def __new__(cls, *args, **kwargs) -> "Formula":
+        """Method for creating a new instance of the class."""
+        result = cls.evaluate(*args, **kwargs)
+        return float.__new__(cls, result)
```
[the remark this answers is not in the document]
SZeltaat: I was just thinking! Could we use __call__ method instead of __new__ method? I know it would be a completely different approach, but it serves the same purpose of "not needing to call the evaluate method to get the result". I'm sure there will be downsides to that, but I can't think of one right now at the end of my work week! 😩 😉 
Please let me know if you need an example to be able to assess my suggestion.

blueprints/codes/formula.py
```diff
@@ -0,0 +1,62 @@
+"""Module for the abstract base class Formula."""
+from abc import ABC, abstractmethod
+
+
+class Formula(float, ABC):
```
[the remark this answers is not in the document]
egarciamendez: Good point @SZeltaat but I think we should stick to the mathematical definition because 'Formula' is supposed to be used on different types of codes and not only on the eurocodes.

blueprints/codes/formula.py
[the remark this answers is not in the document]
egarciamendez: Switching from __new__ to __call__ in Formula means shifting from customizing object creation to defining behavior when an instance is used like a function. This changes usage to a two-step process (create, then call) and is unsuitable for immutable types like float, where value setup happens at instantiation. Essentially, __call__ makes instances function-like, diverging from the instantiation customization achieved with __new__, which is necessary for extending float.

blueprints/codes/formula.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    Abstract base class for formulas used in the codes.
+    """
+
+    def __new__(cls, *args, **kwargs) -> "Formula":
```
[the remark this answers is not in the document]
egarciamendez: afbeelding
That change would return an error.

blueprints/codes/formula.py
```diff
@@ -0,0 +1,62 @@
+"""Module for the abstract base class Formula."""
+from abc import ABC, abstractmethod
```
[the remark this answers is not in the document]
egarciamendez: The 'Formula' class serves as an abstract base class, designed primarily to define an interface for all Formula subclasses within Blueprints. Simultaneously, it behaves similarly to a float. Therefore, annotating it as 'Formula' instead of 'float' in other contexts may not be a desirable situation.

blueprints/codes/formula.py
```diff
@@ -0,0 +1,62 @@
...
+    @property
+    @abstractmethod
+    def source_document(self) -> str:
+        """Property for the source document.
+
+        Returns
+        -------
+        str
+            The reference to the document where the formula originates.
+            This is an abstract method and must be implemented in all subclasses.
```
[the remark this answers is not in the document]
egarciamendez: I have added some examples to the docstrings.

blueprints/codes/formula.py
```diff
@@ -10,7 +10,28 @@ class Formula(float, ABC):
...
-        return float.__new__(cls, result)
+        instance = float.__new__(cls, result)
+        instance._initialized = False
+        return instance
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._initialized = True
+
+    def __setattr__(self, name, value):
```
egarciamendez, line 21: @bro-wi Thanks for the idea.

blueprints/codes/formula.py
```diff
@@ -0,0 +1,62 @@
...
+        Returns
+        -------
+        dict
+            The detailed result of the formula.
+            Keys are strings representing the name of the partial or intermediate result.
+            Values types will depend on the specific implementation, but must be a serializable type.
+        """
+        raise NotImplementedError
+
+    @staticmethod
```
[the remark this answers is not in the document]
johan-tuls: After a second thought, this would enforce strict copying of the original formula. So that helps in consistancy. So static method is fine. However the API of this method is than equal to the init. So i Suggest the docstring of the init is leading and the evaluate only contains the logic to evaluate without a detailed docstring (which is already present in the init)

johan-tuls (approved):

bro-wi (approved):
Good discussions, I'm enjoying it. 

I think this is ready to go! 🚀

Issue #12, opened 2023-11-27
[✨ Feature request]: From 1992-1-1+C2 2011 add formula 7.3


egarciamendez:
### Title

Add the EC formula 7.3 of 1992-1-1+C2 2011: kc coefficient

### Description

Calculate the formula 7.3 for the coefficient kc for flanges of tubular cross-sections and T-sections

### Acceptance Criteria

- Formula 7.3
- Unitesting
- Complete docstrings
- Check reference in README

### Dependencies

None, pure python.

### Confirmation

- [X] I have checked that this feature does not already exist in the issue tracker.
- [X] I have read and understood the contribution guidelines.

PR #13, opened 2023-11-27
Add formula 7.3 from 1992 1 1+c2 2011

egarciamendez:
Check issue #12.

Calculate the formula 7.3 for the coefficient kc for flanges of tubular cross-sections and T-sections

Acceptance Criteria

    Formula 7.3
    Unitesting
    Complete docstrings
    Check reference in README


egarciamendez:
giphy

blueprints/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_7_serviceability_limit_state.py
LBSmulders, line 1: Missing "Returns" in the docstring of "__init__"

blueprints/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_7_serviceability_limit_state.py
```diff
@@ -0,0 +1,49 @@
...
+        self.a_ct = a_ct
+        self.f_ct_eff = f_ct_eff
+
+    @staticmethod
+    def _evaluate(
+        f_cr: KN,
+        a_ct: MM2,
+        f_ct_eff: MPA,
+    ) -> float:
+        """For more detailed documentation see the class docstring."""
```
johan-tuls, line 44: what is the convention we use for the _evaluate method? It is actually part of the init as it is private and activate in the init. However two docstrings are a bit overkill. 
Do we add a default docstring to the def _evaluate method which is always the same? 
`
"""Evaluates the formula, for more infor see the __init__ method"""
`

blueprints/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_7_serviceability_limit_state.py
```diff
@@ -0,0 +1,49 @@
...
+# pylint: disable=arguments-differ
+from blueprints.codes.formula import Formula
+from blueprints.typehints import KN, KN_TO_N, MM2, MPA
+
+
+class Form7Dot3CoefficientKc(Formula):
+    """Class representing the formula 7.3 for the coefficient kc for flanges of tubular cross-sections and T-sections [-]."""
+
+    label = "7.3"
+    source_document = "NEN-EN 1992-1-1+C2:2011"
```
johan-tuls, line 11: could be imported from the init of the NEN-EN 1992-1-1+C2:2011 package. Where it is defined as a constant

blueprints/typehints.py
```diff
@@ -5,3 +5,8 @@
 These constants are intended to improve code readability and prevent the use of 'magic numbers' in calculations.
 Distances are defined with postfixes indicating their units for clarity in usage throughout the code.
 """
+KN = float
+MM2 = float
+MPA = float
+
+KN_TO_N = 1e3
```
johan-tuls, line 12: do we want to place conversions (KN_TO_N) within the typehints.py or in unit_conversion.py?

tests/codes/eurocode/nen_en_1992_1_1_c2_2011/test_chapter_7_serviceability_limit_state.py
```diff
@@ -0,0 +1,60 @@
+"""Testing module for chapter 7 of NEN-EN 1992-1-1+C2:2011."""
```
johan-tuls, line 1: Is there a convention of placing all chapter 7 formula tests in one file? I personally prefer to test eacah object in a single file for a nice overview, also if you run Pytest, you can see which files fail in one eye sight

blueprints/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_7_serviceability_limit_state.py
egarciamendez: This was intended from my part.

tests/codes/test_formula.py
```diff
@@ -0,0 +1,50 @@
+"""Module for testing the Formula class."""
+# pylint: disable=arguments-differ
+import pytest
+
+from blueprints.codes.formula import Formula
+
+
+class DummyTestingFormula(Formula):
```
johan-tuls, line 8: naming TestFormula?

johan-tuls (commented):
Mainly regarding conventions
* One test file per object? or all objects in a single file. Which will probably be really large
* Unit conversions in its own file instead of combined in typehints.py

blueprints/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_7_serviceability_limit_state.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(
+        self,
+        f_cr: KN,
+        a_ct: MM2,
+        f_ct_eff: MPA,
```
bro-wi, line 17: I think we have to settle what units we are using over the complete project--> if we want the formulas to be able to be inside a pipeline we shouldn't have to worry about units. So we can do a few things: 1. agree on standard units in the library and use type aliases 2. use a library like pint for the unit conversions 3. do unit checks ourselves and extend our in and output types with a unit.

I know it is some work, but if we don't do this correctly I think we will have a lot of headache in the future. 

Maybe we should start a private discussion on this?

I would actually suggest to do Newton as input instead of KN --> less unit conversions and it is equal to the input into software like say: idea and scia

blueprints/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_7_serviceability_limit_state.py
bro-wi: I wouldn't even mind if the static evaluate would be public, that gives you the chance to run the function lightweight if you are only interested in a 'float' answer.

About sharing docstrings, maybe we can get some inspiration from this thread: https://softwareengineering.stackexchange.com/questions/386755/sharing-docstrings-between-similar-functions

blueprints/typehints.py
bro-wi: Or no unit conversions at all (stick to SI standard) : https://en.wikipedia.org/wiki/International_System_of_Units

bro-wi (commented):
Nice work! 

My main concern is that we have to agree a bit more on units.

tests/codes/test_formula.py
egarciamendez: Done

blueprints/typehints.py
johan-tuls: During the last meeting we saw that each eurocode has a unit definition at the top which is valid for the entire document. 
We thought it would be best to stick to these units and use these units as typehints, as explicit >> implicit. 

SI standard is really usefull for the back-end, but in a lot of cases this means a lot of conversion/difference with the units written down in the source documents. 

What do you think about this @bro-wi ? My personal preference is to make the unit explicit, eventhough the formula might be unitless, and stick to the unit used in the source document.

blueprints/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_7_serviceability_limit_state.py
johan-tuls: I agree on the private discussion. I'll open one

blueprints/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_7_serviceability_limit_state.py
egarciamendez: Yes, the docstring of the parameters for the formula are already defined in the __init__

blueprints/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_7_serviceability_limit_state.py
egarciamendez: I prefer to keep things simple, but I'm open to initializing it in the document module. What do others think?

blueprints/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_7_serviceability_limit_state.py
egarciamendez: In our last maintainers meeting, we agreed to avoid standard units across the package, opting instead for units that are most commonly used in specific formulas. For instance, we discussed using kNm for bending moments as it's more conventional than Nm or Nmm. Similarly, mm² was preferred over m² for cross-sectional areas. I welcome further discussion on this topic and agree on the importance of getting it right. One essential point I believe we should adhere to is clearly stating the required units in both the type hints and docstrings of each Formula.
I will start a private discussion on this matter.

blueprints/typehints.py
bro-wi: I totally agree. My main concern is how we align formula's as input to a different formula in a intuitive way.

egarciamendez: The ouput unit of the formula is explicit as well so that would solve the problem.

bro-wi: If we continue exploring the "back-trace intermediate calculations" we have to take care of a bit more I believe ;-)

blueprints/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_7_serviceability_limit_state.py
bro-wi: Importing a string a string from a parent folder isn't that complex isn't it? :-p I think it prevents mistakes in the following formulas.

johan-tuls (approved):

PR #39, opened 2023-12-03
(#38) Add a validations.py module

### LINKED ISSUE #38: [✨ Feature request]: Add a validations modules to Blueprints
### Description

Currently all logic to validate formulas is written inside Formula._valuate. We want to separate logic in a separate module puerely for that purpose.

### Acceptance Criteria

- [x] separate validations module

### Dependencies

None

### Confirmation

- [X] I have checked that this feature does not already exist in the issue tracker.
- [X] I have read and understood the contribution guidelines.

egarciamendez:
## Description

Currently all logic to validate formulas is written inside Formula._valuate. We want to separate logic in a separate module puerely for that purpose.

Fixes #38

## Type of change

## Checklist:


blueprints/validations.py
```diff
@@ -0,0 +1,49 @@
...
+    Raise a ValueError if any of the given keyword arguments are negative.
+
+    Parameters
+    ----------
+    **kwargs
+        A dictionary of keyword arguments where keys are parameter names, and values are the values to validate.
+
+    Raises
+    ------
+    ValueError
```
johan-tuls, line 43: NegativeValueError

blueprints/validations.py
```diff
@@ -0,0 +1,49 @@
...
+    Raise a ValueError if any of the given keyword arguments are less than or equal to zero.
+
+    Parameters
+    ----------
+    **kwargs
+        A dictionary of keyword arguments where keys are parameter names, and values are the values to validate.
+
+    Raises
+    ------
+    ValueError
```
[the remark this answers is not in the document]
egarciamendez: 😉

blueprints/validations.py
egarciamendez: 😉

egarciamendez:
Guys can we merge this one?

blueprints/validations.py
```diff
@@ -0,0 +1,49 @@
+"""Module for validation actions inside of Blueprints."""
+
+
+class NonPositiveValueError(Exception):
+    """Raised when a value is less than or equal to zero."""
```
bro-wi, line 5: In my opinion it is quite confusing to use "Litotes" in error messages, but I can't think of a better one so just leave it.

blueprints/validations.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class NegativeValueError(Exception):
+    """Raised when a value is negative."""
```
bro-wi, line 9: Does it add value to add these exceptions to an exception group?

blueprints/validations.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def raise_if_less_or_equal_to_zero(**kwargs) -> None:
```
bro-wi, line 12: question (not specifically about this PR): do we want to force kwargs input for our functions/classes? I'm not sure if our current signature are already in this style: "Formula(self,*,a,b,c). But that would do the trick.

bro-wi (commented):
nice and clean solution!

I only have some minor things to consider, but if you want to move them to a separate issue that is fine by me, in that case you can merge.

blueprints/validations.py
egarciamendez: Maybe then LessOrEqualToZeroError???

blueprints/validations.py
egarciamendez: What do you mean?

blueprints/validations.py
egarciamendez: That's a good point. We can make it more general by requesting a dictionary, rather than enforcing the use of kwargs entirely. The place where this is implemented should then unpack the kwargs first.

```python
def raise_if_negative(values: dict[str, float]) -> None:
    """
    Raise a NegativeValueError if any value in the provided dictionary is negative.

    Parameters
    ----------
    values : dict[str, float]
        A dictionary where keys are parameter names, and values are the values to validate.

    Raises
    ------
    NegativeValueError
        If any value in the dictionary is negative.
    """
    # Iterate over each item in the dictionary
    for key, value in values.items():
        # Check if the value is negative
        if value < 0:
            # If a negative value is found, raise the NegativeValueError with a message
            raise NegativeValueError(f"Invalid value for '{key}': {value}. Values for '{key}' cannot be negative.")
```

blueprints/validations.py
bro-wi: https://realpython.com/python311-exception-groups/

blueprints/validations.py
bro-wi: Mwaah that's not my point, my point is more on the init side ;-)

I think this is more-or-less the same as using **kwargs, the usage is a bit different ofcourse:

`self.raise_if_negative(a=1,b=2,c=3)`

vs 

`self.raise_if_negative({"a":1,"b":2,"c":3})`


I think I like the current implementation (first) better...

blueprints/validations.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Raises
+    ------
+    NonPositiveValueError
+        If any value is greater than zero.
+
+    """
+    for key, value in kwargs.items():
+        if value <= 0:
+            raise NonPositiveValueError(f"Invalid value for '{key}': {value}. Values for '{key}' must be greater than zero.")
```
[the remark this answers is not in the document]
egarciamendez: Done 😉

Issue #41, opened 2023-12-04
[✨ Feature request]: Latex representation of formula


johan-tuls:
### Description

Each formula will, for traceability and verifiability purposes, need someway to represent itself. 

### Acceptance Criteria

- [ ] It should give a Latex representation of the Symbol = Original formula = Numeric formula = Result
- [ ] Symbel only
- [ ] Formula symbolic
- [ ] Formula Numeric
- [ ] Result
- [ ] Adjust decimals

(Nice to have's ? )
- [ ] Show Units for the result only
- [ ] Show Units for each argument
- [ ] Avoid defining logic in different places, this should be the responsibility of the _evaluate method

### Dependencies

Python only, however Latex syntax. So in the future adapters could be written from Latex --> external packages.

### Confirmation

- [X] I have checked that this feature does not already exist in the issue tracker.
- [X] I have read and understood the contribution guidelines.

egarciamendez:
Looking good 🚀 And extra nice to have would be the option to add the formula label at the end.

PR #45, opened 2023-12-07
#41 Added initial version of the LatexFormula

johan-tuls:
## Description

In order to represent the formula in whatever context as desired. 
The formula should have a standard way of representing itself. Most likely in Latex format.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

## Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas


egarciamendez (commented):
I like this aproach but I am very curious to see what @bro-wi is cooking.

I would only like to add a reference to the formula label at the end. And maybe treat this object as a str in the same way we did with float and Formula. But that is more of a suggestion.

johan-tuls:
> I was planning to spend some time next week on this subject, (although my mind is constantly prototyping). So I hope you're not in too much of a hurry ;-)

Not in a hurry ^_^ But i think once this feature has been implemented, the formula class can have really added value and it would be great to start using these formulas outside of blueprints.

bro-wi:
Have been trying to cook something, but it quickly becomes very (and maybe unnecessarily) complex. For a straightforward one-line formula we can make it work by overiding the __add__,__sub__, __mul__,__div__, __pow__ . you have to write some logic to handle adding parenthesis (which quickly becomes quite complex), but it could work. a small (and way to simple) example:

```
@dataclass(frozen=True)
class LatexFormula:
    return_symbol: str
    result: float
    equation: str = ""
    numeric_equation: str = ""

    @property
    def complete(self) -> str:
        all_sub_equations = [self.return_symbol, self.equation, self.numeric_equation, str(self.result)]
        return " = ".join([eq for eq in all_sub_equations if eq != ""])

    @property
    def short(self) -> str:
        return f"{self.return_symbol} = {self.result}"

    def __type_check(self, other):
        if not isinstance(other, LatexFormula):
            raise TypeError(f"can't add LatexFormula to non-LatexFormula of type '{type(other)}'")

    def __str__(self):
        return self.complete

    def __add__(self, other):
        self.__type_check(other)

        return LatexFormula(
            "", self.result + other.result, f"{self.equation} + {other.equation}", f"{self.numeric_equation}+{other.numeric_equation}"
        )

    def __mul__(self, other):
        self.__type_check(other)

        # TODO: write check so parentheses are added when self or other is a summation, but not when there are already parentheses around them:
        # e.g.: a+b * c+d should be (a+b)*(c*d)
        #       (a+b)*c is already okay
        #       a+b * c should be (a+b)*c

        return LatexFormula(
            "", self.result * other.result, f"({self.equation})({other.equation})", f"({self.numeric_equation})({other.numeric_equation})"
        )
```

would work like this:

image


--> but then: 
-what will happen for multiline formulas? 
-And how would we include the formula label?
-Another thing I cant really tackle (yet) is inequality (the current implementation also fails on that).

so for the time being: I think have to KISS. I'm afraid that opting something very complex now will turn out to be more a hassle than we just 'type in' the result we want. It takes more effort of the developers, but I think it is more foolproof and flexible. Sorry for taking so long to come to this conclusion.. there were some escalations that required my attention.

egarciamendez:
Task remaining before merging:

- [ ] Add docstrings
- [ ] Add tests

bro-wi:
> Tasks remaining before merging:
> 
> * [ ]  Add docstrings
> * [ ]  Add tests

*[ ] fix for inequality
*[ ] add formula label

blueprints/codes/latex_formula.py
```diff
@@ -0,0 +1,18 @@
...
+class LatexFormula:
+    return_symbol: str
+    result: str
+    equation: str = ""
+    numeric_equation: str = ""
+
+    @property
+    def complete(self) -> str:
+        all_sub_equations = [self.return_symbol, self.equation, self.numeric_equation, self.result]
+        return " = ".join([eq for eq in all_sub_equations if eq != ""])
```
bro-wi, line 14: this will (probably) not work for inequalities?
+ include formula label

blueprints/codes/latex_formula.py
```diff
@@ -0,0 +1,18 @@
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class LatexFormula:
+    return_symbol: str
+    result: str
+    equation: str = ""
+    numeric_equation: str = ""
```
bro-wi, line 9: add docstrings  (also mention that the units are included in the numeric equation )

johan-tuls:
> Glad we have almost reached a conclusion. @johan-tuls Can you add this to your solution?
> 
> ```python
>     def __str__(self):
>         return self.complete
> ```

Done

blueprints/codes/latex_formula.py
johan-tuls: I have added the formula label as comparison_operator_label with = as default.

johan-tuls:
I have added the open requests. Let's see if this works out for the initial version.
@bro-wi @egarciamendez could you review the changes?

blueprints/codes/latex_formula.py
```diff
@@ -0,0 +1,57 @@
...
+        The formula with symbols
+    numeric_equation: str, default ""
+        The formula with values (numbers)
+    comparison_operator_label: str, default "="
+        The label for the comparison operators between the return symbol and the result.
+        Could be changed for inequalities.
+    """
+
+    return_symbol: str
+    result: str
```
egarciamendez, line 26: I think this should be a float instead. Transforming to a str is the responsibility of the class itself.

blueprints/codes/latex_formula.py
egarciamendez, line 1: What do we do with inequalities? This class doesn't cover those situations.

tests/codes/test_latex_formula.py
```diff
@@ -0,0 +1,39 @@
+"""Tests for the LatexFormula class."""
+from blueprints.codes.latex_formula import LatexFormula
+
+
+class TestLatexFormula:
+    """Test for LatexFormula."""
+
+    @staticmethod
+    def fixture_1() -> LatexFormula:
```
egarciamendez, line 9: Use @pytest.fixtures instead and put the fixture outside of the testing class.

tests/codes/test_latex_formula.py
```diff
@@ -0,0 +1,39 @@
...
+
+        assert formula.short == expected_result
+
+    def test_complete(self) -> None:
+        """Test the complete representation."""
+        # Example values
+        formula = self.fixture_1()
+
+        # Expected result
+        expected_result = "E = E=mc^2 = 5*10^2 = 500"
```
egarciamendez, line 29: The expected result has an extra unnecessary E -> "E = mc^2 =...

egarciamendez (changes requested):
We're almost there 💪

tests/codes/test_latex_formula.py
johan-tuls: Fixed, the initial fixture was wrong

tests/codes/test_latex_formula.py
johan-tuls: done, great improvement!

docs/source/conf.py
```diff
@@ -3,9 +3,10 @@
 import sys
 from datetime import datetime
 
-sys.path.insert(0, os.path.abspath("../../"))
```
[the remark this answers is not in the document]
johan-tuls: reversed

blueprints/codes/latex_formula.py
johan-tuls: as it only contains strings and in theorie should be possible for any formula (including string results) i suggest we stick to str for all subresults.

blueprints/codes/latex_formula.py
johan-tuls: most of them can be rewritten as formulas. Otherwise we might need a new object

egarciamendez (approved):
I think this should be the way to go. 🚀

egarciamendez:
Let's wait for @bro-wi 's review to merge.

blueprints/codes/latex_formula.py
```diff
@@ -0,0 +1,58 @@
...
+        """Complete representation of the formula.
+
+        Returns
+        -------
+        str
+            Return symbol = equation = numeric_equation = result
+
+        """
+        all_sub_equations = [self.return_symbol, self.equation, self.numeric_equation, self.result]
+        return f" {self.comparison_operator_label} ".join([eq for eq in all_sub_equations if eq != ""])
```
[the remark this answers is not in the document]
johan-tuls: Thanks for the review. 
I indeed think inequalities might need a different approach. The problem is that most inequalities aren't really inequalities, but just formula written in a different format so it is more readable for the end-user. 
So we might have to debate what would be the best approach.

However i think that for generic formula (y = .....x ...) we should try to complete the PR so we can start implementing Blueprints in regular apps and giving some momentum to the devlopment.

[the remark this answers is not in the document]
bro-wi: agree! We already have issue #50 to keep the inequalities on our radar, so let's merge and move on 🚀

PR #306, opened 2024-07-08
Refactor Exposure classes and adjust tests

### LINKED ISSUE #305: [✨ Feature request]: ABC Exposure class
### Description

Exposure classes are now defined in a specific version of the Eurocode. The description of the exposure classes and the possibilties in each category can be different in the national annex of each country and are also subject to change in the future versions.

Since other classes will depend on these classes, there should be an abstract exposure class, abstract version of each exposure class and an abstract version of ExposureClasses data container.

This was initially not done, because inheriting from Enum and ABC at the same time is not possible. Here we will apply this stackoverflow solution:
https://stackoverflow.com/questions/56131308/create-an-abstract-enum-class/56135108#56135108 

### Acceptance Criteria

- This feature should be non-breaking

### Dependencies

_No response_

### Confirmation

- [X] I have checked that this feature does not already exist in the issue tracker.
- [X] I have read and understood the contribution guidelines.

SZeltaat:
## Description

Refactored and replaced exposure classes. Main change is the definition of ABCEnumMeta class to be able to make Exposure an abstract class. This was required if we want to define the concept of exposure class independant of the exact definition in different Eurocode versions.
I had to do this in order to be able to do the same for the structural class and the cover calculation.
All the other changes are mostly replacing code.

Please also read #305
Fixes #305 

## Type of change

## Checklist:


tests/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_4_durability_and_cover/test_table_4_1.py
```diff
@@ -185,44 +108,6 @@ def test_description_of_the_environment(self) -> None:
...
-            chloride_seawater=ChlorideSeawater("Not applicable"),
-            freeze=FreezeThaw("XF2"),
-            chemical=Chemical("XA2"),
-        )
-        assert exposureclasses.no_risk is False
-
-
 def test_comparing_different_types_raises_error() -> None:
     """Check if comparing different exposure class types, raises TypeError."""
     with pytest.raises(TypeError):
```
egarciamendez, line 113: To solve the errors regarding 'Statement seems to have no effect'. You could add an underscore as a variable
afbeelding

egarciamendez (approved):
Beautiful !
afbeelding

blueprints/codes/eurocode/exposure_classes.py
```diff
@@ -0,0 +1,166 @@
...
+        Boolean
+            True if both arguments are of the same severity (In this case they will both be literally the same).
+        """
+        if isinstance(other, self.__class__):
+            _self_severity = int(self.value[-1]) if self.value != "Not applicable" else 0
+            _other_severity = int(other.value[-1]) if other.value != "Not applicable" else 0
+            return _self_severity == _other_severity
+        raise TypeError("Only the same exposure class types can be compared with each other!")
+
+    def __gt__(self, other: object) -> bool:
```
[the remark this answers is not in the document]
SZeltaat: Somehow, `other: T` works fine here and mypy doesn't mind that. For the sake of consistency, I will NOT change this one as well.

tests/codes/eurocode/nen_en_1992_1_1_c2_2011/chapter_4_durability_and_cover/test_table_4_1.py
SZeltaat: Done!
I did not get these errors in my IDE though! 🤷‍♂️

SZeltaat (commented):
Thank you for your review!
I tried changing 'object' to 'T', but mypy complains that the signature is different from its superclass. I have tried a couple of other possible solutions, but with no luck. So I'm keeping 'object' in. If you can think of a solution, please let me know or feel free to apply it yourself!
I'll wait for your response before merging! 😉

egarciamendez:
Sometimes is a 9 just fine 👌

PR #332, opened 2024-08-01
Fix wrong TypeError at instantiating MockAbstractEnum

### LINKED ISSUE #327: [🐞 Bug]: Fix test_mock_abstract_enum_is_abstract
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/blueprints/) of Blueprints.

### Issue description

test_mock_abstract_enum_is_abstract catches a TypeError, but the wrong one.
Fix: 

```python
class MockAbstractEnumWithMembers(Enum, metaclass=ABCEnumMeta):
    """Mock class with members for testing the ABCEnumMeta class."""

    A = "a"

    @abstractmethod
    def test_method(self) -> int:
        """Test method for the MockAbstractEnum class."""
```

```python

    def test_mock_abstract_enum_is_abstract(self) -> None:
        """Test case for MockAbstractEnum class to check if it's an abstract class."""
        assert MockAbstractEnum.__abstractmethods__ == frozenset({"test_method"})
        with pytest.raises(TypeError) as error_info:
            _ = MockAbstractEnumWithMembers("a")  # type: ignore[abstract]
        assert str(error_info.value) == "Can't instantiate abstract class 'MockAbstractEnumWithMembers' without an implementation for abstract method 'test_method'"

```

### Reproducible example

The last assert would show that the TypeError in current implementation of the test is not the right one.

```python

    def test_mock_abstract_enum_is_abstract(self) -> None:
        """Test case for MockAbstractEnum class to check if it's an abstract class."""
        assert MockAbstractEnum.__abstractmethods__ == frozenset({"test_method"})
        with pytest.raises(TypeError) as error_info:
            _ = MockAbstractEnumWithMembers("a")  # type: ignore[abstract]
        assert str(error_info.value) == "Can't instantiate abstract class 'MockAbstractEnumWithMembers' without an implementation for abstract method 'test_method'"
```

### Expected behavior

wrong test

### Installed versions

```
latest
```

SZeltaat:
## Description
We've discussed this yesterday. I think this should fix it. We now check if the expected TypeError is raised and not accidentally another TypeError.

Fixes #327 

## Type of change

## Checklist:

### DIFF
--- a/tests/utils/test_abc_enum_meta.py
+++ b/tests/utils/test_abc_enum_meta.py
@@ -38,8 +38,24 @@ class TestABCEnumMeta:
     def test_mock_abstract_enum_is_abstract(self) -> None:
         """Test case for MockAbstractEnum class to check if it's an abstract class."""
         assert MockAbstractEnum.__abstractmethods__ == frozenset({"test_method"})
-        with pytest.raises(TypeError):
-            _ = MockAbstractEnum("a")  # type: ignore[abstract]
+
+    def test_abstract_enum_with_memebers_raises_type_error(self) -> None:
+        """Test case for MockAbstractEnumWithMembers class to check if it raises a TypeError."""
+        with pytest.raises(TypeError) as error_info:
+
+            class MockAbstractEnumWithMembers(Enum, metaclass=ABCEnumMeta):
+                """Mock class with members for testing the ABCEnumMeta class."""
+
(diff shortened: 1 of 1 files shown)


egarciamendez (approved):
👍

PR #546, opened 2025-04-04
feat: Add standard steel profiles various shapes

GerjanDorgelo:
part 2 of #535 

- [x] This change requires a documentation update

## Checklist:


GerjanDorgelo:
@SZeltaat I added you for review. I think easiest way to check all this is to make a small script/excel that checks the sizes of all elements based on your own database. This PR is part of a large (huge?) PR, see #535. That one will also do some math with the standard profiles like calculate area, section moduli, etc.

blueprints/structural_sections/steel/steel_cross_sections/standard_profiles/unp.py
```diff
@@ -0,0 +1,47 @@
...
+    def __init__(self, code: str, h: MM, b: MM, t_w: MM, t_f: MM, radius: MM, slope: RATIO) -> None:
+        """Initialize UNP profile.
+        Args:
+            code (str): Profile code.
+            h (MM): Total height.
+            b (MM): Total width.
+            t_w (MM): Web thickness.
+            t_f (MM): Flange thickness.
+            radius (MM): Radius (always 0 for UNP profiles).
+            slope (RATIO): Slope of the flange.
```
SZeltaat, line 37: this does not make sense to me! radius is always zero, but it has value for all the enum members. and what is the slope actually? Maybe look at this again to check if everything is ok?

tests/structural_sections/steel/steel_cross_sections/standard_profiles/test_hea.py
```diff
@@ -0,0 +1,35 @@
...
+
+class TestHEA:
+    """Tests for the HEA enum."""
+
+    def test_enum_values(self) -> None:
+        """Test that enum values are correctly defined."""
+        assert HEA.HEA_100.value == ("HEA100", 96, 100, 5, 8, 12)
+        assert HEA.HEA_200.value == ("HEA200", 190, 200, 6.5, 10, 18)
+
+    def test_enum_membership(self) -> None:
```
SZeltaat, line 14: random check is of course fine, but maybe it's nice to check for the first and the last enum member.

SZeltaat (changes requested):
Nice addition!

I mostly have comments that are valid for all the sections. also for the tests. So please apply them where applicable. I've tried to be as explicit as I could, but I have probably missed some along the way.

tests/structural_sections/steel/steel_cross_sections/standard_profiles/test_hea.py
GerjanDorgelo: good point

GerjanDorgelo:
> Nice addition!
> 
> I mostly have comments that are valid for all the sections. also for the tests. So please apply them where applicable. I've tried to be as explicit as I could, but I have probably missed some along the way.

Thank you! I implemented all your suggestions, except for:
- The sources, i used various or combined multiple table bases. There doesnt seem to be a single source of truth for the profile shapes.
- implementing height of web: as their appear to be various definitions of "web height", as specified in the review item specifically

Could you give this another look @SZeltaat ? Thank youu

SZeltaat (approved):
Great job,
Thank you for yet another valuable contribution! 😉

PR #651, opened 2025-07-20
Update the list of examples

teknovizier:
## Description

Update the example list to include all currently available examples.


bro-wi (approved):
Great addition, thanks for adding the missing examples in the overview list! :-)

PR #652, opened 2025-07-20
Update `README.md` with minimum required Python version

teknovizier:
## Description

Mention in `README.md` that Python >=3.12 is required.


bro-wi (approved):
Good improvement, this is more explicit than the icon alone.

PR #691, opened 2025-07-26
657 feature request add light test for faster local test runs

### LINKED ISSUE #657: [✨ Feature request]: Add light-test for faster local test runs
### Description

Currently, there are a couple of tests that take too long to run and that makes running the tests locally cumbersome.

The idea is to add mocked versions of those tests and have a light test command in the Makefile that ignores the slow tests. We still keep the slow tests and will run them by default in the CI/CD and those can be run locally by the current make commands.

### Acceptance Criteria

- [ ] Identify the slow tests
- [ ] add mocked version of slow tests
- [ ] add light-test command to Makefile

### Dependencies

_No response_

### Confirmation

- [x] I have checked that this feature does not already exist in the issue tracker.
- [x] I have read and understood the contribution guidelines.

SZeltaat:
## Description

First of all, >90% of the changes concern uv.lock, so this is a small PR 😅 

In this PR:
- [x] I've added pytest-xdist as a test dependency to run tests in parallel. This has increased speed by ~35% (on my machine 🙄).
- [x] I've also included the parallel execution of the tests in the CI/CD.
- [x] I've marked 4 slow tests as slow and added "make test-light" with these slow tests excluded. 🐌 
- [x] I've improved the performance of the plot (which was included in the slow tests) function by 10x. Slow tests now take between 1 and 3 seconds. 🐇
- [x] I've added mocked version of slow tests so that "make test-light" also tests the slow tests but with `section_properties` mocked.

Fixes #657 

## Type of change

## Checklist:

### DIFF
--- a/blueprints/structural_sections/steel/steel_cross_sections/plotters/general_steel_plotter.py
+++ b/blueprints/structural_sections/steel/steel_cross_sections/plotters/general_steel_plotter.py
@@ -56,12 +56,13 @@ def plot_shapes(
     ax.plot(profile.centroid.x, profile.centroid.y, "o", color="black")
 
     # Add legend text
+    profile_section_properties = profile.section_properties(plastic=False, warping=False)
     legend_text = f"""
     {profile.name}\n
     Area: {profile.area:.1f} mm²
     Weight per meter: {profile.weight_per_meter:.1f} kg/m
-    Moment of inertia about x: {profile.section_properties().ixx_c:.0f} mm⁴
-    Moment of inertia about y: {profile.section_properties().iyy_c:.0f} mm⁴
+    Moment of inertia about x: {profile_section_properties.ixx_c:.0f} mm⁴
+    Moment of inertia about y: {profile_section_properties.iyy_c:.0f} mm⁴
     """
 
     # Add the steel quality if all elements have the same material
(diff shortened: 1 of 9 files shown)


blueprints/structural_sections/steel/steel_cross_sections/plotters/general_steel_plotter.py
```diff
@@ -56,12 +56,13 @@ def plot_shapes(
     ax.plot(profile.centroid.x, profile.centroid.y, "o", color="black")
 
     # Add legend text
+    profile_section_properties = profile.section_properties(plastic=False, warping=False)
```
bro-wi, line 59: Off topic: should we add a cache decorator to section properties? :) maybe we can open a new issue for that.

bro-wi (approved):
Very nice, and good catch on the plotting code. That is very inefficient indeed!

egarciamendez (approved):

blueprints/structural_sections/steel/steel_cross_sections/plotters/general_steel_plotter.py
egarciamendez: Good idea. I think we should.

PR #754, opened 2025-09-16
Polygonbuilder: implement public API

### LINKED ISSUE #746: [Feature✨] CrossSectionBuilder: implement interface

SZeltaat:
## Description

This is the first step into implementing the PolygonBuilder. In this step, I've added the public API.
In the following PR, I will add the logic.

Fixes #746 

## Type of change

## Checklist:

### DIFF
--- a/blueprints/structural_sections/_polygon_builder.py
+++ b/blueprints/structural_sections/_polygon_builder.py
@@ -61,20 +61,13 @@ class PolygonBuilder:
         self._points: NDArray[np.float64] = np.empty((0, 2), dtype=float)
         self._current: NDArray[np.float64] | None = None
 
-    def add_line(self, start: PointLike, length: Length, angle: DEG) -> PolygonBuilder:
-        """Add a straight line segment to the polygon.
-
-        This method sets the starting point of the polygon if it is not already set.
+    def set_starting_point(self, start: PointLike) -> PolygonBuilder:
+        """Set the starting vertex for the polygon being constructed.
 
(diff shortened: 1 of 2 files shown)


egarciamendez:
This methods looks and does almost the same. What can be confusing.
add_arc/append_arc and add_line/append_line

SZeltaat:
I see!

I needed both:
Append: for not repreating the end points and increasing the chance of floating point conflicts
Add: for the case where we want to introduce holes in the section. I haven't add that to the api yet, but That is something we would need for hollow sections.

However, what do you think of the following?

- Maybe we can remove the `add` methods and introduce a `set_starting_point` method. Then we will always use the `append` methods. It is the most Pythonic name after all!
- We then handle the holes in another manner. Maybe even after creating the polygons (which now seems the logical way to go).

egarciamendez:
> I see!
> 
> 
> 
> I needed both:
> 
> Append: for not repreating the end points and increasing the chance of floating point conflicts
> 
> Add: for the case where we want to introduce holes in the section. I haven't add that to the api yet, but That is something we would need for hollow sections.
> 
> 
> 
> However, what do you think of the following?
> 
> 
> 
> - Maybe we can remove the `add` methods and introduce a `set_starting_point` method. Then we will always use the `append` methods. It is the most Pythonic name after all!
> 
> - We then handle the holes in another manner. Maybe even after creating the polygons (which now seems the logical way to go).

Sounds like a good plan 👏

PR #791, opened 2025-10-09
Ec3addformula_Table 5.2 Member subject to bending and compression with test

MichaelOcampo1104:
## Description

Added Table 5.2 Part subject  to bending and compression with test

## Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works
- [-] I have commented my code, particularly in hard-to-understand areas
- [-] I have made corresponding changes to the documentation
- [-] New and existing unit tests pass locally with my changes


GerjanDorgelo:
i kinda like the approach that has been used in this PR. when looking at the table.. its not really a table, its just formatted as is. Basically table 5.1 part 1 is just 3 (series of) equations using c, t, alpha and fy as input variables, one of which is featured in this PR. It would need clarification that it only does the Internal compression parts though, and im not sure where the current equations for alpha and psi come from

MichaelOcampo1104:
@GerjanDorgelo , Yes, the way I approach Table 5.2 Sheet 1/3 is provide it with 3 equation instead of a formal table., this approach can apply similarly for Sheet 2 and 3, as for the alpha and psi , these values are derived from the equilibrium of stress distribution as illustrated in Table 5.2 Sheet1/3

PR #879, opened 2025-12-24
861 refactor standard profiles should subclass profile

### LINKED ISSUE #861: [Refactor 🛠️] standard profiles should subclass Profile

SZeltaat:
## Description

Hi @egarciamendez. I have already discussed this with Wichard. The approach with Enums had a huge overhead that created all the standard profiles in our library, twice even if we only needed one standard profile. That is due to the nature of Enums. In addition, I was overriding the Enum Meta which is potentially a risk of unknown behaviour. I have tried 2 alternatives, 1: using descriptors (which had the same disadvantage of initiating all profiles eagerly on class definition) and 2: class getattr (this PR).

This approach has a slight disadvantage of creating the stubs, but I can accept that because:
- We only need to do it once (very rarely we might want to add new standard profiles!)
- It can be automated (I have written a script that does it automatically for all standard profiles) or done by copliot/AI.

This way we have all the advantages of user-friendly API and clean Pythonic code, without unnecessary overhead at runtime.

Could you please review this PR and let me know what you think?
If approved, I'll make the changes to all the other standard profiles.

Fixes #861 

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Checklist:

- [ ] I have made corresponding changes to the documentation


blueprints/structural_sections/steel/standard_profiles/utils.py
```diff
@@ -43,3 +45,42 @@ def wrapper(self: S, *args: P.args, **kwargs: P.kwargs) -> R:
         return wrapper
 
     return decorator
+
+
+class StandardProfileProtocol(Protocol):
+    """Protocol for standard profile classes."""
+
+    _factory: Callable[..., Profile]
```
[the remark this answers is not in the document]
SZeltaat: nitpick

blueprints/structural_sections/steel/profile_definitions/chs_profile.py
```diff
@@ -66,44 +62,40 @@ def _polygon(self) -> Polygon:
...
-        adjusted_thickness = profile.thickness - corrosion_outside - corrosion_inside
+        adjusted_outer_diameter = self.outer_diameter - 2 * corrosion_outside
+        adjusted_thickness = self.wall_thickness - corrosion_outside - corrosion_inside
 
         if adjusted_thickness <= 0:
             raise ValueError("The profile has fully corroded.")
 
-        name = profile.alias
+        name = self.name
         if corrosion_inside or corrosion_outside:
```
bro-wi: not a really big deal (and probably not used in practice), but if with_corossion is used multiple times (chained) then it will result in multiple repetitions of this suffix correct? --> the benefit is that the 'history' is traceable.

blueprints/structural_sections/steel/standard_profiles/chs.py
```diff
@@ -1,293 +1,286 @@
...
-from __future__ import annotations
-
-from enum import Enum
-
 from blueprints.structural_sections.steel.profile_definitions.chs_profile import CHSProfile
-from blueprints.structural_sections.steel.standard_profiles.utils import wrap_as_instance_method
-from blueprints.type_alias import MM
+from blueprints.structural_sections.steel.standard_profiles.utils import StandardProfileMeta
 
+CHS_PROFILES = {
```
bro-wi: Maybe rename this, CHS_PROFILES indicates that it stores actual profiles, while this only stores the parameters. Maybe even consider a named tuple for the parameters. I think the named tuple would also be benificial for maintaining the keyword/value pairs in one place in stead of in CHS AND keeping the correct order here.

blueprints/structural_sections/steel/profile_definitions/chs_profile.py
SZeltaat: I have solved this somewhere else. 😉

blueprints/structural_sections/steel/standard_profiles/chs.py
```diff
@@ -1,293 +1,286 @@
...
-            Outer diameter of the CHS.
-        thickness : MM
-            Wall thickness of the CHS.
-        """
-        self.alias = alias
-        self.diameter = diameter
-        self.thickness = thickness
+    Usage example
+    -------------
+        >>> profile = CHS.CHS21_3x2_3
```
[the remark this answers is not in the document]
SZeltaat: As discussed, we'll keep it like this, accepting the fact that the user should import one class and get an object of another class. It is a factory after all 😉

blueprints/structural_sections/steel/standard_profiles/chs.py
SZeltaat: As discussed, I have used a namedtuple here and adjusted the metaclass to take that into account.

SZeltaat (commented):
❤️

PR #1090, opened 2026-07-25
feat: add formulas 8.33, 8.34 and 8.35 from FprEN 1992-1-1:2023 clause 8.2.2

egarciamendez:
## Description

Implements Formulas (8.33), (8.34) and (8.35) of clause 8.2.2(5), the three quantities that Formula (8.32) combines for the approach that considers the effect of compressive normal forces:

- `Form8Dot33ShearStressResistanceWithoutAxialForce` for (8.33), the shear stress resistance before the axial force term.
- `Form8Dot34FactorK1` for (8.34), the factor by which the normal stress is weighted.
- `Form8Dot35MaximumShearStressResistance` for (8.35), the upper bound used in (8.32).

Formula (8.32) itself follows in a separate pull request, once these three exist to refer to.

Points worth a reviewer's attention:

- **The upper bound in (8.34) is not a plain 0.18.** It carries the same `A_c / (b_w * z)` factor as the expression, so for the geometry used in the tests the bound is 0.24, not 0.18. Clipping at 0.18 would look reasonable and be wrong; there is a test for the case where the bound governs.
- **`e_p` is signed and may be negative.** The standard defines it as positive towards the tensile side, so an eccentricity on the compression side is a valid input. It has no sign guard. The standard also prints no lower bound on `k_1`, so an eccentricity below `-d/3` yields a negative factor. Both are tested rather than suppressed.
- (8.33) is the expression of (8.27) without its lower bound. It is implemented separately rather than shared, because (8.27) is still in review in #1088. Once that lands, folding (8.27) into `max(Form8Dot33(...), tau_rdc_min)` would remove the duplication, and I am happy to do that as a follow up if the reviewer prefers it.
- Both printed bounds are implemented as a minimum, matching the treatment of (8.27) and (8.30).
- The class name for (8.33) deserves a second opinion. The standard calls `tau_Rdc,0` simply "the design value of the shear stress resistance", but that name is already taken by (8.27), so a qualifier was needed. If `WithoutAxialForce` reads wrong, say so and I will rename it.

Contributes to #1083

## Formulas as implemented

**Formula (8.33)**, `Form8Dot33ShearStressResistanceWithoutAxialForce`

`equation`

$$
\tau_{Rdc,0} = \frac{0.66}{\gamma_V} \cdot \left(100 \cdot \rho_l \cdot f_{ck} \cdot \frac{d_{dg}}{d}\right)^{\frac{1}{3}}
$$

`complete`

$$
[...]

Issue #1083, opened 2026-07-24
[🔧 Formula Request]: Add formulas from chapter 8 from Eurocode FprEN 1992-1-1:2023


egarciamendez:
## Formula Description

Add the 129 formulas of chapter 8 (Ultimate Limit States) from FprEN 1992-1-1:2023, numbered (8.1) through (8.129), to the project.

This is an umbrella issue. The work is delivered clause by clause, at most three formula classes per pull request, so that each review can be checked against a single page of the standard. Pull requests reference this issue with `Contributes to #NNN` and do not close it.

## Acceptance Criteria

- [ ] Add the formula to the code
- [ ] Add latex expression to the formula
- [ ] Add unittests
- [ ] Add docstrings
- [ ] Update table in docs with classname(s)

## Progress by clause

A clause is ticked once its pull request is merged, not when it is opened.

### 8.2.1 General verification procedure

- [x] (3) and (4): formulas (8.18), (8.19), (8.20), merged in #1084
- [x] (5): formulas (8.21), (8.22), (8.23), (8.24), merged in #1085
- [x] (5): formulas (8.25), (8.26), merged in #1086

### 8.2.2 Detailed verification for members without shear reinforcement

- [x] (2): formulas (8.27), (8.28), merged in #1088
- [x] (3) and (4): formulas (8.29), (8.30), (8.31), merged in #1089
- [x] (5): formulas (8.33), (8.34), (8.35), merged in #1090
- [x] (5): formula (8.32), merged in #1091
- [x] (6): formulas (8.36), (8.37), merged in #1092
- [x] (7): formulas (8.38), (8.39), (8.40), merged in #1094

### 8.2.3 Members with shear reinforcement

- [x] (4) to (9): formulas (8.41) to (8.53), merged in #1095 by @GerjanDorgelo. Our own (8.41) in #1096 was closed in favour of it.
- [x] (10) and (12): formulas (8.54), (8.55), merged in #1098
- [x] (12): formulas (8.56), (8.57), merged in #1099
- [x] (13): formulas (8.58), (8.59), (8.60), merged in #1100
- [x] (13): formulas (8.61), (8.62), (8.63), merged in #1101

### 8.2.4 In-plane shear and transverse bending

- [ ] (2): formula (8.64), open in #1102

### 8.2.5 Shear between web and flanges

- [ ] (1) and (2): formulas (8.65), (8.66), open in #1103
- [ ] (3): formulas (8.67), (8.68), open in #1106
[...]

PR #1091, opened 2026-07-25
feat: add formula 8.32 from FprEN 1992-1-1:2023 clause 8.2.2

egarciamendez:
## Description

Implements Formula (8.32) of clause 8.2.2(5) as `Form8Dot32DesignShearStressResistanceWithNormalForce`. This is the assembly that the three quantities of #1090 feed: the design value of the shear stress resistance for the approach that considers the effect of compressive normal forces, offered by the standard as an alternative to Formula (8.27) combined with Formula (8.31).

Points worth a reviewer's attention:

- **The standard prints this as a chained inequality with an embedded equality**, `tau_Rdc,min <= tau_Rd,c = tau_Rdc,0 - k_1 * sigma_cp <= tau_Rdc,max`, so strictly it reads as a verification rather than an assignment. It is implemented as a clamp, first against the lower bound and then against the upper one, which is what the bounds are for and matches the treatment of (8.27) and (8.30). All three outcomes have a test.
- **The operator between the terms is a minus.** With `sigma_cp = N_Ed / A_c` and compression entering negative, compression raises the resistance, which is the point of the clause. Implementing it as a plus while feeding a positive compressive stress produces the same number for compression and the wrong sign for tension.
- **Neither `k_1` nor `sigma_cp` has a sign guard, deliberately.** Formula (8.34) can produce a negative `k_1`, and `sigma_cp` is negative in compression and positive in tension. Restricting either would break the formula.
- The clamping order only matters for the degenerate input where the lower bound exceeds the upper one. The standard prints no rule for that, so the behaviour is documented in the docstring rather than rejected by a guard the standard does not ask for.
- `tau_rdc_0`, `tau_rdc_min` and `tau_rdc_max` come in as parameters rather than being derived internally, following the pattern used throughout this clause. The class therefore cannot check that `tau_Rdc,max <= 2.7 * tau_Rdc,0`, which Formula (8.35) guarantees at its own level.
- Minor rendering observation, not specific to this formula: a negative substituted value renders as `0.133 \cdot -2.000`, a multiplication dot directly followed by a minus sign. Wrapping negative values in parentheses would read better, but that belongs in `latex_replace_symbols` and would affect every formula in the library, so it is left alone here.

The docstring refers to the classes of Formulas (8.33), (8.34) and (8.35), which are in review in #1090. There is no code dependency: this pull request takes their results as parameters and can be merged independently.

Contributes to #1083

## Formulas as implemented

**Formula (8.32)**, `Form8Dot32DesignShearStressResistanceWithNormalForce`

`equation`

$$
\tau_{Rd,c} = \min\left(\max\left(\tau_{Rdc,0} - k_1 \cdot \sigma_{cp}, \tau_{Rdc,min}\right), \tau_{Rdc,max}\right)
$$

`complete`

$$
\tau_{Rd,c} = \min\left(\max\left(\tau_{Rdc,0} - k_1 \cdot \sigma_{cp}, \tau_{Rdc,min}\right), \tau_{Rdc,max}\right) = \min\left(\max\left(0.586 - 0.133 \cdot -2.000, 0.522\right), 1.483\right) = 0.853 \ MPa
$$

`complete_with_units`

$$
\tau_{Rd,c} = \min\left(\max\left(\tau_{Rdc,0} - k_1 \cdot \sigma_{cp}, \tau_{Rdc,min}\right), \tau_{Rdc,max}\right) = \min\left(\max\left(0.586 \ MPa - 0.133 \cdot -2.000 \ MPa, 0.522 \ MPa\right), 1.483 \ MPa\right) = 0.853 \ MPa
$$

## Type of change

## Checklist

- [x] Tests for the unclamped case, both bounds, and a negative factor with a result that stays between the bounds so the clamping cannot hide the sign handling
- [x] Docstrings added, with variable descriptions taken from the standard
- [x] Documentation table updated with the class name
- [x] `blueprints check` passes: lint, format, type check, and 100% coverage

## Summary by CodeRabbit

* **New Features**
  * Added Eurocode 2 formula **8.32** for design shear stress resistance with compressive normal force.
  * Resistance is clamped between configurable minimum and maximum limits.
  * Supports intermediate (unclamped) and final (clamped) result reporting, plus symbolic/numeric/unit-aware LaTeX output.
* **Documentation**
  * Updated the Eurocode formula guide to mark **8.32** as implemented.
* **Tests**
  * Added automated tests covering evaluation, clamping behavior, validation errors (including inverted bounds), intermediate vs final results, and LaTeX formatting.

### DIFF
--- a/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
+++ b/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
@@ -82,6 +82,17 @@ class Form8Dot32DesignShearStressResistanceWithNormalForce(Formula):
         """Evaluates the formula, for more information see the __init__ method."""
         raise_if_negative(tau_rdc_0=tau_rdc_0, tau_rdc_min=tau_rdc_min, tau_rdc_max=tau_rdc_max)
 
+        # The two bounds come from different formulas, (8.20) and (8.35), and nothing ties them together.
+        # (8.20) does not depend on the reinforcement ratio while (8.35) does, so for a lightly reinforced
+        # member the maximum can fall below the minimum and the printed chain has no solution at all. Clamping
+        # anyway would return one of the two bounds and quietly break the relation this formula prints, so the
+        # combination is refused instead. The standard says nothing about it; this is a decision taken here.
+        if tau_rdc_min > tau_rdc_max:
+            raise ValueError(
+                f"The minimum shear stress resistance of {tau_rdc_min} MPa exceeds the maximum of {tau_rdc_max} MPa, "
+                f"so no value can satisfy Formula (8.32). Check the inputs of Formulas (8.20) and (8.35)."
+            )
+
         return min(max(tau_rdc_0 - k_1 * sigma_cp, tau_rdc_min), tau_rdc_max)
 
     def latex(self, n: int = 3) -> LatexFormula:
--- a/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
+++ b/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
@@ -42,10 +42,21 @@ class Form8Dot32DesignShearStressResistanceWithNormalForce(Formula):
             normal forces according to Formula (8.33), see Form8Dot33ShearStressResistanceWithoutAxialForce [$MPa$].
         k_1 : DIMENSIONLESS
             [$k_1$] Factor according to Formula (8.34), see Form8Dot34FactorK1, unless the National Annex gives
-            another value. The standard prints no lower bound on it, so a negative value is valid [$-$].
+            another value. It is not required to be positive and is not guarded here: Formula (8.34) prints only
+            an upper bound, and its eccentricity [$e_p$] is defined as positive towards the tensile side, so a
+            tendon eccentric towards the compression side gives a negative [$e_p$] and can drive [$k_1$] below
+            zero [$-$].
         sigma_cp : MPA
             [$\sigma_{cp}$] Normal stress, defined by the standard as [$N_{Ed} / A_c$] with [$A_c$] the area of
-            the concrete cross-section. Its sign carries meaning and is used as given [$MPa$].
+            the concrete cross-section. **Compression is negative.** The standard does not say so anywhere near
+            this formula, but the printed minus sign settles it: compression has to raise the shear stress
+            resistance, and with a positive [$k_1$] the term [$- k_1 \cdot \sigma_{cp}$] only does that for a
+            negative [$\sigma_{cp}$]. It also follows from [$N_{Ed}$] itself, which this clause treats as
+            positive in tension, as Formula (8.31) does.
+
+            Note that the sign of this term flipped between the two generations of the code. EN 1992-1-1:2004
+            prints [$+ k_1 \cdot \sigma_{cp}$] with compression positive, so a value carried over from a
+            calculation to that code has the wrong sign here [$MPa$].
         tau_rdc_min : MPA
             [$\tau_{Rdc,min}$] Minimum shear stress resistance according to Formula (8.20), see
             Form8Dot20MinimumShearStressResistance [$MPa$].
@@ -81,7 +92,7 @@ class Form8Dot32DesignShearStressResistanceWithNormalForce(Formula):
             replacements={
                 r"\tau_{Rdc,0}": f"{self.tau_rdc_0:.{n}f}",
                 r"k_1": f"{self.k_1:.{n}f}",
-                r"\sigma_{cp}": f"{self.sigma_cp:.{n}f}",
+                r"\sigma_{cp}": r"\left(" + f"{self.sigma_cp:.{n}f}" + r"\right)",
                 r"\tau_{Rdc,min}": f"{self.tau_rdc_min:.{n}f}",
                 r"\tau_{Rdc,max}": f"{self.tau_rdc_max:.{n}f}",
             },
@@ -92,15 +103,20 @@ class Form8Dot32DesignShearStressResistanceWithNormalForce(Formula):
             replacements={
                 r"\tau_{Rdc,0}": rf"{self.tau_rdc_0:.{n}f} \ MPa",
                 r"k_1": f"{self.k_1:.{n}f}",
-                r"\sigma_{cp}": rf"{self.sigma_cp:.{n}f} \ MPa",
+                r"\sigma_{cp}": r"\left(" + rf"{self.sigma_cp:.{n}f} \ MPa" + r"\right)",
                 r"\tau_{Rdc,min}": rf"{self.tau_rdc_min:.{n}f} \ MPa",
                 r"\tau_{Rdc,max}": rf"{self.tau_rdc_max:.{n}f} \ MPa",
             },
             unique_symbol_check=False,
         )
+        # The value of the expression before the two bounds are applied. The line above already shows the
+        # arithmetic, so what this adds is whether a bound was active: it differs from the result exactly then.
+        _intermediate: str = f"{self.tau_rdc_0 - self.k_1 * self.sigma_cp:.{n}f}"
+
         return LatexFormula(
             return_symbol=r"\tau_{Rd,c}",
             result=f"{self:.{n}f}",
+            intermediate_result=_intermediate,
             equation=_equation,
             numeric_equation=_numeric_equation,
             numeric_equation_with_units=_numeric_equation_with_units,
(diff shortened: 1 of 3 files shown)


blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
```diff
@@ -0,0 +1,109 @@
...
+    @staticmethod
+    def _evaluate(
+        tau_rdc_0: MPA,
+        k_1: DIMENSIONLESS,
+        sigma_cp: MPA,
+        tau_rdc_min: MPA,
+        tau_rdc_max: MPA,
+    ) -> MPA:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(tau_rdc_0=tau_rdc_0, tau_rdc_min=tau_rdc_min, tau_rdc_max=tau_rdc_max)
```
rhhk-dot, line 83: Factor `k_1` should also be non-negative.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
```diff
@@ -0,0 +1,109 @@
...
+
+        return min(max(tau_rdc_0 - k_1 * sigma_cp, tau_rdc_min), tau_rdc_max)
```
rhhk-dot, line 96: It seems strange that `k_1 * sigma_cp` lowers the value of `tau_rdc`, assuming a positive value of `sigma_cp` means compression. In the current Eurocode it is the other way around. I did not find any reference to a sign convention near this formula / clause which indicates what sign corresponds with compression or tension.

Image

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
```diff
@@ -0,0 +1,109 @@
...
+        ----------
+        tau_rdc_0 : MPA
+            [$\tau_{Rdc,0}$] Design value of the shear stress resistance without the effect of compressive
+            normal forces according to Formula (8.33), see Form8Dot33ShearStressResistanceWithoutAxialForce [$MPa$].
+        k_1 : DIMENSIONLESS
+            [$k_1$] Factor according to Formula (8.34), see Form8Dot34FactorK1, unless the National Annex gives
+            another value. The standard prints no lower bound on it, so a negative value is valid [$-$].
+        sigma_cp : MPA
+            [$\sigma_{cp}$] Normal stress, defined by the standard as [$N_{Ed} / A_c$] with [$A_c$] the area of
+            the concrete cross-section. Its sign carries meaning and is used as given [$MPa$].
```
rhhk-dot: Define what the sign means. In the current implementation a negative value corresponds with compression, but that is not clearly mentioned in the Eurocode.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
```diff
@@ -0,0 +1,109 @@
...
+        sigma_cp: MPA,
+        tau_rdc_min: MPA,
+        tau_rdc_max: MPA,
+    ) -> MPA:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(tau_rdc_0=tau_rdc_0, tau_rdc_min=tau_rdc_min, tau_rdc_max=tau_rdc_max)
+
+        return min(max(tau_rdc_0 - k_1 * sigma_cp, tau_rdc_min), tau_rdc_max)
+
+    def latex(self, n: int = 3) -> LatexFormula:
```
rhhk-dot, line 98: Adding an intermediate result would be nice.

rhhk-dot (changes requested):
Checked

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
egarciamendez: Documented in 6e2699a8: compression is negative.

The standard does not say it near this formula, but the printed minus sign settles it. Compression has to raise the resistance, and with a positive `k_1` the term `- k_1 * sigma_cp` only does that for a negative `sigma_cp`. It also follows from `N_Ed`, which this clause treats as positive in tension, as (8.31) does.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
egarciamendez: Your reasoning is right, the assumption is what differs: here a positive `sigma_cp` is tension, not compression. So `- k_1 * sigma_cp` raises the resistance under compression, as it should.

You are also right that EN 1992-1-1:2004 has it the other way: `+ k_1 * sigma_cp` with compression positive. The sign of the term flipped between the two generations. That is now in the docstring, including the warning that a value carried over from a 2004 calculation has the wrong sign here.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
egarciamendez: Added. It shows the value of the expression before clamping, so it differs from the result exactly when one of the two bounds bit:

`... = min(max(0.586 - 0.133 * (2.000), 0.400), 1.500) = 0.320 = 0.400 MPa`

Also bracketed the normal stress, since a negative value is the ordinary case here and `0.133 * -2.000` read badly.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
egarciamendez: Not doing this one, because it would refuse input that Formula (8.34) can legitimately produce.

`k_1` comes from (8.34), which prints only an upper bound. Its eccentricity `e_p` is defined by the standard as positive towards the tensile side, so a tendon eccentric towards the compression side gives a negative `e_p`, and `0.5/a_cs,0 * (e_p + d/3)` can then go below zero. Guarding `k_1 >= 0` here would mean two formulas from the same clause contradict each other.

The blind verifier of (8.34) reached the same reading independently in #1090 and was explicit that clamping `k_1` at zero would be an addition to the standard.

The rationale is now in the docstring and in `test_evaluation_with_a_negative_factor`. Say if you read (8.34) differently.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_32.py
```diff
@@ -0,0 +1,125 @@
...
+        tau_rdc_0: MPA,
+        k_1: DIMENSIONLESS,
+        sigma_cp: MPA,
+        tau_rdc_min: MPA,
+        tau_rdc_max: MPA,
+    ) -> MPA:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(tau_rdc_0=tau_rdc_0, tau_rdc_min=tau_rdc_min, tau_rdc_max=tau_rdc_max)
+
+        return min(max(tau_rdc_0 - k_1 * sigma_cp, tau_rdc_min), tau_rdc_max)
```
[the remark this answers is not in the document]
egarciamendez: Valid, and reachable with realistic input rather than only degenerate input. Fixed in f8c5497b: the combination now raises.

The two bounds come from different formulas and nothing ties them together. The minimum of (8.20) does not depend on the reinforcement ratio, the maximum of (8.35) does through (8.33), so they cross for a lightly reinforced member. With gamma_v = 1.4, f_ck = 30, f_yd = 435, d_dg = 32, d = 500 and a_cs_0 = 1333.333:

```
tau_Rdc,min = 0.5220 MPa  (independent of rho_l)

   rho_l  tau_Rdc,0  tau_Rdc,max   inverted?
    0.01     0.5859       1.4835   no
   0.001     0.2720       0.6886   no
  0.0005     0.2159       0.5465   no
  0.0004     0.2004       0.5073   yes
  0.0002     0.1590       0.4027   yes
```

Refusing is a decision taken here, not a reading: the standard says nothing about this case. But the alternatives are worse. Returning `tau_rdc_max` hands back a value below the declared minimum and breaks the chain the formula prints, and returning `tau_rdc_min` picks the other side just as arbitrarily. The code comment and the regression test both say so, and the test uses the 0.04 percent row above.

PR #1009, opened 2026-03-29
1008 Add formulas NB.NB.2, 7, 8, 9, 11 and 12 from Eurocode 1993-1-1 NB NL

GerjanDorgelo:
## Description

Add formulas NB.NB.2, 7, 8, 9, 11 and 12 from Eurocode 1993-1-1 NB NL

Fixes #1008

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] New and existing unit tests pass locally with my changes

## Release Notes

* **New Features**
  * Added six Dutch National Annex formulas for steel beam buckling design, covering critical buckling moment, reduction factors, parameter S, coefficient C, and alpha.
  * Added support for sixth-power length units in engineering calculations.

* **Documentation**
  * Updated Eurocode formula documentation to show these calculations as available.

* **Tests**
  * Added coverage for numerical calculations, input validation, boundary conditions, and formula rendering.

PR #1092, opened 2026-07-25
feat: add formulas 8.36 and 8.37 from FprEN 1992-1-1:2023 clause 8.2.2

egarciamendez:
## Description

Implements Formulas (8.36) and (8.37) of clause 8.2.2(6), the effective depth and the reinforcement ratio for prestressed members with bonded tendons:

- `Form8Dot36EffectiveDepthPrestressedMembers` for (8.36).
- `Form8Dot37ReinforcementRatioPrestressedMembers` for (8.37).

With this, clause 8.2.2 is complete up to and including paragraph (6).

Points worth a reviewer's attention:

- **Only the depths are squared, and only in the numerator of (8.36).** The areas are never squared, the denominator is linear in the depths, and in (8.37) the single squared quantity is `d` in the denominator. Carrying the squares of (8.36) into (8.37) is the plausible misreading here.
- **The omission allowed by 8.2.2(6) is expressed by passing `a_p = 0`.** The standard permits leaving out the prestressed reinforcement if including it lowers the shear resistance through the reduced effective depth. With `a_p = 0`, (8.36) returns `d_s`, and (8.37) reduces to `A_s / (b_w * d_s)`, which is exactly Formula (8.28) for the same section. There is a test for that identity, which cross-checks the two clauses against each other.
- **`d` in (8.37) is redundant but kept as an input.** Within this clause it is not free: for the same member it follows from (8.36) using the four reinforcement quantities that are already in the signature, so a caller passing something else gets a silently inconsistent ratio. It is kept as a parameter for consistency with the rest of this clause, where derived quantities are passed in rather than computed internally, and the constraint is spelled out in the docstring. Deriving it inside (8.37) and dropping the parameter is the alternative; say the word and I will change it.
- The denominator of (8.36) is guarded as a whole rather than per term, since it is the product sum that must be non-zero. A section with neither ordinary nor prestressed reinforcement raises rather than dividing by zero.

Contributes to #1083

## Formulas as implemented

**Formula (8.36)**, `Form8Dot36EffectiveDepthPrestressedMembers`

`equation`

$$
d = \frac{\left(d_s\right)^2 \cdot A_s + \left(d_p\right)^2 \cdot A_p}{d_s \cdot A_s + d_p \cdot A_p}
$$

`complete`

$$
d = \frac{\left(d_s\right)^2 \cdot A_s + \left(d_p\right)^2 \cdot A_p}{d_s \cdot A_s + d_p \cdot A_p} = \frac{\left(500.000\right)^2 \cdot 1500.000 + \left(450.000\right)^2 \cdot 800.000}{500.000 \cdot 1500.000 + 450.000 \cdot 800.000} = 483.784 \ mm
$$

`complete_with_units`

$$
d = \frac{\left(d_s\right)^2 \cdot A_s + \left(d_p\right)^2 \cdot A_p}{d_s \cdot A_s + d_p \cdot A_p} = \frac{\left(500.000 \ mm\right)^2 \cdot 1500.000 \ mm^2 + \left(450.000 \ mm\right)^2 \cdot 800.000 \ mm^2}{500.000 \ mm \cdot 1500.000 \ mm^2 + 450.000 \ mm \cdot 800.000 \ mm^2} = 483.784 \ mm
$$

**Formula (8.37)**, `Form8Dot37ReinforcementRatioPrestressedMembers`

`equation`

$$
\rho_l = \frac{d_s \cdot A_s + d_p \cdot A_p}{b_w \cdot \left(d\right)^2}
$$

`complete`

$$
\rho_l = \frac{d_s \cdot A_s + d_p \cdot A_p}{b_w \cdot \left(d\right)^2} = \frac{500.000 \cdot 1500.000 + 450.000 \cdot 800.000}{300.000 \cdot \left(483.784\right)^2} = 0.016 \ -
$$

`complete_with_units`

$$
\rho_l = \frac{d_s \cdot A_s + d_p \cdot A_p}{b_w \cdot \left(d\right)^2} = \frac{500.000 \ mm \cdot 1500.000 \ mm^2 + 450.000 \ mm \cdot 800.000 \ mm^2}{300.000 \ mm \cdot \left(483.784 \ mm\right)^2} = 0.016 \ -
$$

## Type of change

## Checklist

- [x] Tests for the general case, the omission of the prestressed reinforcement, the identity with Formula (8.28), and every raise
- [x] Docstrings added, with variable descriptions taken from the standard
- [x] Documentation table updated with the class names
- [x] `blueprints check` passes: lint, format, type check, and 100% coverage

## Summary by CodeRabbit

* **New Features**
  * Added Eurocode 2 (2023) Formula **8.36** for effective depth of prestressed members with bonded tendons.
  * Added Eurocode 2 (2023) Formula **8.37** for reinforcement ratio of prestressed members, including input validation and correct reductions when prestressing is omitted.
  * Added symbolic and numeric **LaTeX** outputs (with and without units) for both formulas.
* **Documentation**
  * Updated the Eurocode formula guide to mark **8.36** and **8.37** as completed and link the corresponding formula names.
* **Tests**
  * Added tests covering calculations, validation/edge cases, and LaTeX formatting.

### DIFF
--- a/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_36.py
+++ b/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_36.py
@@ -1,0 +1,86 @@
+"""Formula 8.36 from FprEN 1992-1-1:2023: Chapter 8: Ultimate limit states (ULS)."""
+
+from blueprints.codes.eurocode.fpr_en_1992_1_1_2023 import FPR_EN_1992_1_1_2023
+from blueprints.codes.formula import Formula
+from blueprints.codes.latex_formula import LatexFormula, latex_replace_symbols
+from blueprints.type_alias import MM, MM2
+from blueprints.validations import raise_if_less_or_equal_to_zero, raise_if_negative
+
+
+class Form8Dot36EffectiveDepthPrestressedMembers(Formula):
+    """Class representing formula 8.36 for the calculation of the effective depth of prestressed members
+    with bonded tendons.
+    """
+
+    label = "8.36"
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(self, d_s: MM, a_s: MM2, d_p: MM, a_p: MM2) -> None:
+        r"""[$d$] Effective depth of prestressed members with bonded tendons [$mm$].
+
+        FprEN 1992-1-1:2023 (E) art 8.2.2 (6) - Formula (8.36)
+
+        The area of prestressed reinforcement [$A_p$] may be omitted if including it reduces the shear
+        resistance through the reduced effective depth, provided the longitudinal tension reinforcement
+        [$A_s$] is sufficient to carry [$M_{Ed}$] and [$N_{Ed}$] taking into account the effect of
+        prestressing. Passing [$A_p = 0$] returns [$d_s$], which is that omission.
+
+        Parameters
+        ----------
+        d_s : MM
+            [$d_s$] Effective depth of the longitudinal tension reinforcement [$mm$].
+        a_s : MM2
+            [$A_s$] Area of the longitudinal tension reinforcement [$mm^2$].
+        d_p : MM
+            [$d_p$] Effective depth of the prestressed reinforcement [$mm$].
+        a_p : MM2
+            [$A_p$] Area of the prestressed reinforcement [$mm^2$].
+        """
+        super().__init__()
+        self.d_s = d_s
+        self.a_s = a_s
+        self.d_p = d_p
+        self.a_p = a_p
+
+    @staticmethod
+    def _evaluate(d_s: MM, a_s: MM2, d_p: MM, a_p: MM2) -> MM:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(d_s=d_s, a_s=a_s, d_p=d_p, a_p=a_p)
+
+        denominator = d_s * a_s + d_p * a_p
+        raise_if_less_or_equal_to_zero(denominator=denominator)
+
+        return (d_s**2 * a_s + d_p**2 * a_p) / denominator
+
+    def latex(self, n: int = 3) -> LatexFormula:
+        """Returns LatexFormula object for formula 8.36."""
+        _equation: str = r"\frac{\left(d_s\right)^2 \cdot A_s + \left(d_p\right)^2 \cdot A_p}{d_s \cdot A_s + d_p \cdot A_p}"
+        _numeric_equation: str = latex_replace_symbols(
+            template=_equation,
+            replacements={
+                r"d_s": f"{self.d_s:.{n}f}",
+                r"A_s": f"{self.a_s:.{n}f}",
+                r"d_p": f"{self.d_p:.{n}f}",
+                r"A_p": f"{self.a_p:.{n}f}",
+            },
+            unique_symbol_check=False,
+        )
+        _numeric_equation_with_units: str = latex_replace_symbols(
+            template=_equation,
+            replacements={
+                r"d_s": rf"{self.d_s:.{n}f} \ mm",
+                r"A_s": rf"{self.a_s:.{n}f} \ mm^2",
+                r"d_p": rf"{self.d_p:.{n}f} \ mm",
+                r"A_p": rf"{self.a_p:.{n}f} \ mm^2",
+            },
+            unique_symbol_check=False,
+        )
+        return LatexFormula(
+            return_symbol=r"d",
+            result=f"{self:.{n}f}",
+            equation=_equation,
+            numeric_equation=_numeric_equation,
+            numeric_equation_with_units=_numeric_equation_with_units,
+            comparison_operator_label="=",
+            unit="mm",
+        )
--- a/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_37.py
+++ b/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_37.py
@@ -1,0 +1,95 @@
+"""Formula 8.37 from FprEN 1992-1-1:2023: Chapter 8: Ultimate limit states (ULS)."""
+
+from blueprints.codes.eurocode.fpr_en_1992_1_1_2023 import FPR_EN_1992_1_1_2023
+from blueprints.codes.formula import Formula
+from blueprints.codes.latex_formula import LatexFormula, latex_replace_symbols
+from blueprints.type_alias import DIMENSIONLESS, MM, MM2
+from blueprints.validations import raise_if_less_or_equal_to_zero, raise_if_negative
+
+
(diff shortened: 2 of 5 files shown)


blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_36.py
```diff
@@ -0,0 +1,86 @@
...
+        raise_if_negative(d_s=d_s, a_s=a_s, d_p=d_p, a_p=a_p)
+
+        denominator = d_s * a_s + d_p * a_p
+        raise_if_less_or_equal_to_zero(denominator=denominator)
+
+        return (d_s**2 * a_s + d_p**2 * a_p) / denominator
+
+    def latex(self, n: int = 3) -> LatexFormula:
+        """Returns LatexFormula object for formula 8.36."""
+        _equation: str = r"\frac{\left(d_s\right)^2 \cdot A_s + \left(d_p\right)^2 \cdot A_p}{d_s \cdot A_s + d_p \cdot A_p}"
```
casper-van-der-krift, line 57: Mathematically, we don't need parentheses around d_s and d_p, before taking the square. Also, it is inconsistent with Eurocode notation.

Image

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_37.py
```diff
@@ -0,0 +1,95 @@
...
+            unique_symbol_check=False,
+        )
+        return LatexFormula(
+            return_symbol=r"\rho_l",
+            result=f"{self:.{n}f}",
+            equation=_equation,
+            numeric_equation=_numeric_equation,
+            numeric_equation_with_units=_numeric_equation_with_units,
+            comparison_operator_label="=",
+            unit="-",
```
casper-van-der-krift, line 94: I find it rather unnatural to return an actual '-' as unit in the LatexFormula instance. Either remove (my preference) or replace by placing it between square brackets '[-]'.

tests/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/test_formula_8_36.py
```diff
@@ -0,0 +1,88 @@
...
+
+        assert formula == pytest.approx(expected=manually_calculated_result, rel=1e-4)
+
+    def test_evaluation_when_the_prestressed_reinforcement_is_omitted(self) -> None:
+        """Tests the omission of the prestressed reinforcement allowed by 8.2.2(6), which returns the effective depth
+        of the longitudinal tension reinforcement.
+        """
+        formula = Form8Dot36EffectiveDepthPrestressedMembers(d_s=500.0, a_s=1500.0, d_p=450.0, a_p=0.0)
+
+        assert formula == pytest.approx(expected=500.0, rel=1e-9)
```
casper-van-der-krift: A relative tolerance of 1e-9 is too strict. I don't see why testing without prestress means a decreased relative tolerance. Please change to blueprints default `rel=1e-4`.

tests/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/test_formula_8_37.py
```diff
@@ -0,0 +1,97 @@
...
+    def test_evaluation_reduces_to_formula_8_28_without_prestressing(self) -> None:
+        """Tests that omitting the prestressed reinforcement reproduces Formula (8.28).
+
+        With a_p = 0 the formula becomes d_s * A_s / (b_w * d_s^2), which is A_s / (b_w * d_s), the
+        reinforcement ratio of Formula (8.28) for the same section.
+        """
+        formula = Form8Dot37ReinforcementRatioPrestressedMembers(d_s=500.0, a_s=1500.0, d_p=450.0, a_p=0.0, b_w=300.0, d=500.0)
+
+        # 1500 / (300 * 500), the value Formula (8.28) gives for the same section
+        assert formula == pytest.approx(expected=0.01, rel=1e-9)
```
casper-van-der-krift: SEC

casper-van-der-krift (changes requested):

tests/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/test_formula_8_36.py
egarciamendez: Changed, but to exact equality rather than `1e-4`.

You are right that "without prestress" is no reason for a tighter tolerance. The actual reason was that the case is algebraically exact: with `A_p = 0` the formula reduces to `d_s^2 * A_s / (d_s * A_s)` = `d_s`, and in floating point that comes out as exactly 500.0. `rel=1e-9` said that badly, `rel=1e-4` would drop the claim entirely, so it is now `assert float(formula) == 500.0` with the reduction spelled out in the docstring.

Say if you would rather have the repo default for consistency.

tests/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/test_formula_8_37.py
egarciamendez: Read this as the same point on the 8.37 side, so the same change: `assert float(formula) == 0.01`, since `A_s / (b_w * d_s)` is exact there too. Correct me if you meant something else.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_36.py
egarciamendez: Keeping them, but not for the maths. You are right on both counts you raise: they are mathematically redundant and the standard prints `d_s^2` bare. They are there for the value substitution, which happens by plain text replacement.

Without them, `complete_with_units` renders:

```
\frac{500.000 \ mm^2 \cdot 1500.000 \ mm^2 + ...
```

The exponent attaches to the unit instead of to the value, so `d_s^2` reads as five hundred square millimetres, right next to `1500.000 mm^2` which really is an area. With them:

```
\frac{\left(500.000 \ mm\right)^2 \cdot 1500.000 \ mm^2 + ...
```

`equation` and `complete` stay correct either way, which is what makes it easy to miss. This exact defect was found in #1009, in two of the six formulas of the Dutch national annex to EN 1993-1-1, and is a rule in the skill since.

The alternative would be dropping units from `complete_with_units` for squared quantities, which seems worse.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_37.py
egarciamendez: Not changing it here, but I think you are right in general.

`unit="-"` is in 41 files on main across several Eurocodes, so 8.37 would become the only exception. Worth a separate issue for the whole repository rather than one formula; `[-]` or an empty unit would both read better than the bare minus.

PR #1085, opened 2026-07-25
feat: add formulas 8.21 to 8.24 from FprEN 1992-1-1:2023 clause 8.2.1

egarciamendez:
## Description

Implements Formulas (8.21) to (8.24) of clause 8.2.1(5), for planar members with out-of-plane shear forces acting in two directions:

- `Form8Dot21DesignShearForcePerUnitWidth` for (8.21), the resultant design shear force per unit width.
- `Form8Dot22To24EffectiveDepth` for (8.22), (8.23) and (8.24), the effective depth as a function of the ratio of the shear forces. The three numbered formulas are branches of a single rule, so they share one class, following the pattern of `Form6Dot58And59TensileForce`. All three rows in the documentation table point at it.

Points worth a reviewer's attention:

- The branch boundaries in (8.22) to (8.24) are one-sided as printed. A ratio of exactly 0.5 falls under (8.22) and a ratio of exactly 2 falls under (8.24). Both have their own test case.
- The standard gives no rule for `v_Ed,x = 0`, where the ratio is undefined, and none for negative shear forces, where a negative ratio would land in (8.22) unintentionally. Both are rejected by validation and the choice is stated in the docstring.
- The two shear force components are taken as magnitudes in (8.21) as well, so the API is consistent across the clause.

The package rename of #1082 is merged, and this branch is rebased onto `main`. It does not depend on #1084.

Contributes to #1083

## Formulas as implemented

**Formula (8.21)**, `Form8Dot21DesignShearForcePerUnitWidth`

`equation`

$$
v_{Ed} = \sqrt{\left(v_{Ed,x}\right)^2 + \left(v_{Ed,y}\right)^2}
$$

`complete`

$$
v_{Ed} = \sqrt{\left(v_{Ed,x}\right)^2 + \left(v_{Ed,y}\right)^2} = \sqrt{\left(100.000\right)^2 + \left(40.000\right)^2} = 107.703 \ N/mm
$$

`complete_with_units`

$$
v_{Ed} = \sqrt{\left(v_{Ed,x}\right)^2 + \left(v_{Ed,y}\right)^2} = \sqrt{\left(100.000 \ N/mm\right)^2 + \left(40.000 \ N/mm\right)^2} = 107.703 \ N/mm
$$

**Formula (8.22/8.23/8.24)**, `Form8Dot22To24EffectiveDepth`

`equation`

$$
[...]

PR #1094, opened 2026-07-27
feat: add formulas 8.38, 8.39 and 8.40 from FprEN 1992-1-1:2023 clause 8.2.2

egarciamendez:
Implements formulas (8.38), (8.39) and (8.40) of FprEN 1992-1-1:2023 (E), clause 8.2.2(7), page 121.

The three numbered formulas are the three branches of one rule: the reinforcement ratio of a planar member with different reinforcement ratios in both directions, selected on the ratio $v_{Ed,y}/v_{Ed,x}$. They are implemented as one class, `Form8Dot38To40ReinforcementRatioPlanarMembers`, in the same way as (8.22), (8.23) and (8.24) in #1085.

Contributes to #1083.

## Decisions

- The branch boundaries are one-sided exactly as printed: a ratio of exactly 0,5 falls under (8.38), a ratio of exactly 2 falls under (8.40), and (8.39) is open at both ends. The three branches are discontinuous at both boundaries, so this is not cosmetic. Both boundaries have their own test.
- The standard gives no rule for $v_{Ed,x} = 0$, where the ratio is undefined, so that input is rejected. Same decision as in #1085 for (8.22) to (8.24). A reviewer who prefers the other reading, that the ratio tends to infinity and formula (8.40) applies, should say so once and it will be changed in both places.
- $\alpha_v$ is an input, not derived from the shear forces. The standard says only that it is defined in 8.2.1(5), where formula (8.26) gives it as $\arctan(v_{Ed,y}/v_{Ed,x})$ but with "may". The class therefore does not check $\alpha_v$ against the shear forces passed alongside it, and the standard does not ask it to. Same treatment as (8.25) in #1086.
- The standard prints $\cos^4\alpha_v$ bare; the LaTeX writes `\cos^4(\alpha_v)` with parentheses, which is what makes value substitution possible. Same as (8.25).

## Verification

Blindly verified against the rendered pages 118 to 121 of the standard by a second agent that never saw the implementation or the tests.

Verdict **OK**. It independently derived clause 8.2.2(7), the same branch boundaries with the same inclusive sides, the exponent on the trigonometric function rather than on the angle, and the same result for all six test input sets. The two points it raised are the two listed above: $v_{Ed,x} = 0$ is left undefined by the standard, and $\alpha_v$ cannot be checked for consistency from these pages.

## Formulas as implemented

One class covers all three numbered formulas, so there is one set of representations. The values shown are $v_{Ed,x} = 100$ N/mm, $v_{Ed,y} = 57{,}735$ N/mm, $\rho_{l,x} = 0{,}012$, $\rho_{l,y} = 0{,}008$ and $\alpha_v = 30$ degrees, which puts the ratio at $\tan 30^\circ$ and therefore in the middle branch, formula (8.39).

**Formulas (8.38), (8.39) and (8.40)**, `Form8Dot38To40ReinforcementRatioPlanarMembers`

`equation`

$$
\rho_l = \begin{cases} \rho_{l,x} & \text{if } \frac{v_{Ed,y}}{v_{Ed,x}} \leq 0.5 \\ \rho_{l,x} \cdot \cos^4(\alpha_v) + \rho_{l,y} \cdot \sin^4(\alpha_v) & \text{if } 0.5 < \frac{v_{Ed,y}}{v_{Ed,x}} < 2 \\ \rho_{l,y} & \text{if } \frac{v_{Ed,y}}{v_{Ed,x}} \geq 2 \end{cases}
$$

`complete`

$$
\rho_l = \begin{cases} \rho_{l,x} & \text{if } \frac{v_{Ed,y}}{v_{Ed,x}} \leq 0.5 \\ \rho_{l,x} \cdot \cos^4(\alpha_v) + \rho_{l,y} \cdot \sin^4(\alpha_v) & \text{if } 0.5 < \frac{v_{Ed,y}}{v_{Ed,x}} < 2 \\ \rho_{l,y} & \text{if } \frac{v_{Ed,y}}{v_{Ed,x}} \geq 2 \end{cases} = \begin{cases} 0.012 & \text{if } \frac{57.735}{100.000} \leq 0.5 \\ 0.012 \cdot \cos^4(30.000) + 0.008 \cdot \sin^4(30.000) & \text{if } 0.5 < \frac{57.735}{100.000} < 2 \\ 0.008 & \text{if } \frac{57.735}{100.000} \geq 2 \end{cases} = 0.007 \ -
$$

`complete_with_units`

$$
\rho_l = \begin{cases} \rho_{l,x} & \text{if } \frac{v_{Ed,y}}{v_{Ed,x}} \leq 0.5 \\ \rho_{l,x} \cdot \cos^4(\alpha_v) + \rho_{l,y} \cdot \sin^4(\alpha_v) & \text{if } 0.5 < \frac{v_{Ed,y}}{v_{Ed,x}} < 2 \\ \rho_{l,y} & \text{if } \frac{v_{Ed,y}}{v_{Ed,x}} \geq 2 \end{cases} = \begin{cases} 0.012 & \text{if } \frac{57.735 \ N/mm}{100.000 \ N/mm} \leq 0.5 \\ 0.012 \cdot \cos^4(30.000 \ degrees) + 0.008 \cdot \sin^4(30.000 \ degrees) & \text{if } 0.5 < \frac{57.735 \ N/mm}{100.000 \ N/mm} < 2 \\ 0.008 & \text{if } \frac{57.735 \ N/mm}{100.000 \ N/mm} \geq 2 \end{cases} = 0.007 \ -
$$

The other two branches, with the same reinforcement ratios:

| Formula | $v_{Ed,y}/v_{Ed,x}$ | $\alpha_v$ | `short` |
|---|---|---|---|
| (8.38) | 0,400 | 21,801 degrees | $\rho_l = 0.012 \ -$ |
| (8.39) | 0,577 | 30,000 degrees | $\rho_l = 0.007 \ -$ |
| (8.40) | 2,500 | 68,199 degrees | $\rho_l = 0.008 \ -$ |

The lone minus sign after the result is how this repository renders a dimensionless unit; it is left as it is.

## Summary by CodeRabbit

* **New Features**
  * Added Eurocode 2 formulas **8.38–8.40** to compute reinforcement ratios for planar members from the shear-force component ratio.
  * Enhanced formula **8.25** to accept an optional principal shear angle, or derive it from provided shear forces.
* **Documentation**
  * Updated the Eurocode formula status table to mark **8.38–8.40** as implemented.
* **Tests**
  * Added extensive coverage for reinforcement-ratio piecewise branches, angle derivation, input validation, and LaTeX output.

### DIFF
--- a/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_38_39_40.py
+++ b/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_38_39_40.py
@@ -1,0 +1,108 @@
+"""Formula 8.38, 8.39 and 8.40 from FprEN 1992-1-1:2023: Chapter 8: Ultimate limit states (ULS)."""
+
+import numpy as np
+
+from blueprints.codes.eurocode.fpr_en_1992_1_1_2023 import FPR_EN_1992_1_1_2023
+from blueprints.codes.formula import Formula
+from blueprints.codes.latex_formula import LatexFormula, latex_replace_symbols
+from blueprints.type_alias import DEG, DIMENSIONLESS, N_MM
+from blueprints.validations import raise_if_less_or_equal_to_zero, raise_if_negative
+
+
+class Form8Dot38To40ReinforcementRatioPlanarMembers(Formula):
+    """Class representing formulas 8.38, 8.39 and 8.40 for the calculation of the reinforcement ratio of planar members
+    with different reinforcement ratios in both directions, taken as a function of the ratio of the shear forces.
+    """
+
+    label = "8.38/8.39/8.40"
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(self, v_ed_x: N_MM, v_ed_y: N_MM, rho_l_x: DIMENSIONLESS, rho_l_y: DIMENSIONLESS, alpha_v: DEG) -> None:
+        r"""[$\rho_l$] Reinforcement ratio of planar members, as a function of the ratio of the shear forces
+        [$v_{Ed,y}/v_{Ed,x}$] [$-$].
+
+        FprEN 1992-1-1:2023 (E) art 8.2.2 (7) - Formula (8.38), (8.39) and (8.40)
+
+        The boundaries of the ratio are one-sided as printed: a ratio of exactly 0.5 falls under Formula (8.38)
+        and a ratio of exactly 2 falls under Formula (8.40). The standard gives no rule for [$v_{Ed,x} = 0$],
+        for which the ratio is undefined, so that input is rejected. Both shear forces are taken as magnitudes,
+        since a negative ratio would fall under Formula (8.38) without the standard intending it.
+
+        Parameters
+        ----------
+        v_ed_x : N_MM
+            [$v_{Ed,x}$] Out-of-plane design shear force per unit width acting on the cross-section perpendicular
+            to the x direction [$N/mm$].
+        v_ed_y : N_MM
+            [$v_{Ed,y}$] Out-of-plane design shear force per unit width acting on the cross-section perpendicular
+            to the y direction [$N/mm$].
+        rho_l_x : DIMENSIONLESS
+            [$\rho_{l,x}$] Reinforcement ratio in the x direction [$-$].
+        rho_l_y : DIMENSIONLESS
+            [$\rho_{l,y}$] Reinforcement ratio in the y direction [$-$].
+        alpha_v : DEG
+            [$\alpha_v$] Angle between the principal shear force and the x-axis as defined in 8.2.1(5), which may be
+            taken according to Formula (8.26), see Form8Dot26AngleBetweenPrincipalShearForceAndXAxis. It is only used
+            by Formula (8.39), and it is the caller's responsibility to keep it consistent with the shear forces
+            passed here [$degrees$].
+        """
+        super().__init__()
+        self.v_ed_x = v_ed_x
+        self.v_ed_y = v_ed_y
+        self.rho_l_x = rho_l_x
+        self.rho_l_y = rho_l_y
+        self.alpha_v = alpha_v
+
+    @staticmethod
+    def _evaluate(v_ed_x: N_MM, v_ed_y: N_MM, rho_l_x: DIMENSIONLESS, rho_l_y: DIMENSIONLESS, alpha_v: DEG) -> DIMENSIONLESS:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(v_ed_y=v_ed_y, rho_l_x=rho_l_x, rho_l_y=rho_l_y, alpha_v=alpha_v)
+        raise_if_less_or_equal_to_zero(v_ed_x=v_ed_x)
+
+        ratio = v_ed_y / v_ed_x
+        if ratio <= 0.5:
+            return rho_l_x
+        if ratio < 2:
+            alpha_v_rad = np.deg2rad(alpha_v)
+            return rho_l_x * np.cos(alpha_v_rad) ** 4 + rho_l_y * np.sin(alpha_v_rad) ** 4
+        return rho_l_y
+
+    def latex(self, n: int = 3) -> LatexFormula:
+        """Returns LatexFormula object for formulas 8.38, 8.39 and 8.40."""
+        _equation: str = (
+            r"\begin{cases} \rho_{l,x} & \text{if } \frac{v_{Ed,y}}{v_{Ed,x}} \leq 0.5 \\ "
+            r"\rho_{l,x} \cdot \cos^4(\alpha_v) + \rho_{l,y} \cdot \sin^4(\alpha_v) "
+            r"& \text{if } 0.5 < \frac{v_{Ed,y}}{v_{Ed,x}} < 2 \\ "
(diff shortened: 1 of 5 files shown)


blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_38_39_40.py
```diff
@@ -0,0 +1,108 @@
...
+        self.v_ed_x = v_ed_x
+        self.v_ed_y = v_ed_y
+        self.rho_l_x = rho_l_x
+        self.rho_l_y = rho_l_y
+        self.alpha_v = alpha_v
+
+    @staticmethod
+    def _evaluate(v_ed_x: N_MM, v_ed_y: N_MM, rho_l_x: DIMENSIONLESS, rho_l_y: DIMENSIONLESS, alpha_v: DEG) -> DIMENSIONLESS:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(v_ed_y=v_ed_y, rho_l_x=rho_l_x, rho_l_y=rho_l_y, alpha_v=alpha_v)
```
rhhk-dot: Let `v_ed_x` and `v_ed_y` also be negative and work with absolute values. Refer to PR #1085, in which I made a similar comment.

rhhk-dot: `alpha_v` is between 0 and 90 deg, refer to formula 8.26

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_38_39_40.py
```diff
@@ -0,0 +1,108 @@
...
+        raise_if_less_or_equal_to_zero(v_ed_x=v_ed_x)
```
rhhk-dot: See comment regarding `v_ed_x` on R59

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_38_39_40.py
```diff
@@ -0,0 +1,108 @@
...
+
+class Form8Dot38To40ReinforcementRatioPlanarMembers(Formula):
+    """Class representing formulas 8.38, 8.39 and 8.40 for the calculation of the reinforcement ratio of planar members
+    with different reinforcement ratios in both directions, taken as a function of the ratio of the shear forces.
+    """
+
+    label = "8.38/8.39/8.40"
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(self, v_ed_x: N_MM, v_ed_y: N_MM, rho_l_x: DIMENSIONLESS, rho_l_y: DIMENSIONLESS, alpha_v: DEG) -> None:
```
rhhk-dot: Consider making `alpha_v` optional and calculate with formula 8.26 if not provided by the caller. Then the usage of `alpha_v` and corresponding shear forces will be consistent.

rhhk-dot (changes requested):
Checked

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_38_39_40.py
egarciamendez: Done. Both forces are now taken as magnitudes in the branch condition, same as #1085. Added a test that all four sign combinations of _ed_x/_ed_y give the same result.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_38_39_40.py
egarciamendez: Done, see the reply above. The LaTeX now carries the absolute value bars as well.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_38_39_40.py
egarciamendez: Added 
aise_if_greater_than_90, next to the existing 
aise_if_negative. Formula 8.26 is an arctangent of two magnitudes, so it can only return [0, 90) degrees.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_38_39_40.py
egarciamendez: Done. `alpha_v` is now optional and defaults to formula 8.26 on the magnitudes of the two shear forces.

Scope warning: I applied the same to `formula_8_25.py`, which is already on main. There `alpha_v` was mandatory while the shear forces were absent altogether, so it could not default to 8.26 without new arguments. It now takes optional `v_ed_x`/`v_ed_y` and raises `ValueError` if neither route is given. Say the word if you would rather have that split off into its own PR.

Both classes keep an explicit `alpha_v` winning over the computed one: 8.2.1(5) prints 8.26 as what the angle "may be taken as", so a directly determined principal shear direction stays admissible.

PR #1096, opened 2026-07-27
feat: add formula 8.41 from FprEN 1992-1-1:2023 clause 8.2.3

egarciamendez:
Implements formula (8.41) of FprEN 1992-1-1:2023 (E), clause 8.2.3(4), page 122. This opens clause 8.2.3, members with shear reinforcement.

Contributes to #1083.

## A two-sided check built as an aggregation

The standard prints `1 <= cot(theta) <= cot(theta_min)` and introduces it with: "The inclination of the compression field in the web carrying shear may be selected within the following range". The designer selects `cot(theta)`; the formula says whether that selection is admissible. There is no expression on the right of an equals sign, so nothing is being assigned. That is a check, not a value, and it is why (8.32), (8.34) and (8.35) in the earlier pull requests are plain `Formula` instead: those print an assignment with bounds, so they clamp a computed value.

`DoubleComparisonFormula` is on its way out, so the range is built with `AggregatedComparisonFormula`. The one printed relation becomes two comparisons, `SubForm8Dot41LowerBound` and `SubForm8Dot41UpperBound`, joined with `all`.

That gives each bound its own unity check, and the aggregate reports the larger of the two, so the governing bound is visible:

| `cot(theta)` against `cot(theta_min) = 2,5` | verdict | `unity_check` | governed by |
|---|---|---|---|
| 2,0 | OK | 0,80 | upper bound |
| 0,8 | Not OK | 1,25 | lower bound |
| 3,0 | Not OK | 1,20 | upper bound |

Two consequences worth knowing before this pattern spreads through the rest of the standard:

- `lhs` and `rhs` raise `NotImplementedError` on the aggregate by design, so the two halves are reached through `comparison_formulas`. A test pins down all four values.
- `__new__` has to translate the two arguments of this formula into the `aggregation` and `comparison_formulas` pair the base class evaluates. Overriding `_evaluate` instead does not work: its signature is fixed by the base class and `ty` rejects the narrowed override. Every two-sided range in the standard will have to repeat that translation, so a constructor on the base class for this shape would be worth considering.

## Decisions

- Both bounds are inclusive as printed, and both have their own test.
- `cot(theta_min)` is an input, not a separate class. The standard gives three rules for it, 2,5 for ordinary reinforced members, 3,0 under significant axial compression, and `2,5 - 0,1 * N_Ed / |V_Ed| >= 1,0` under axial tension, plus a 20 % reduction for ductility class A. None of the three carries a formula number, so none of them gets a numbered formula class.
- No input validation. The check itself is the answer for any input: a `cot(theta)` below 1 or above `cot(theta_min)` reports Not OK, which is the correct report rather than an error. A `cot(theta_min)` below 1 leaves an empty range, so every selection reports Not OK; there is a test for that. The alternative reading, that such an input should raise, is defensible and a reviewer who prefers it should say so.
- The name says cotangent on purpose. What is checked is `cot(theta)` and not `theta`. The cotangent decreases over the angles that apply here, so the same check written in angles reverses direction. The class docstring says this as well.
- The standard prints `cot(theta)` bare; the LaTeX writes `\cot(\theta)` with parentheses. Without them the substitution key `\cot\theta` is a prefix of `\cot\theta_{min}` and the two would only stay apart by dictionary ordering.

## Verification

Blindly verified against the rendered page 122 of the standard by a second agent that never saw the implementation or the tests.

Verdict **OK**. It independently derived clause 8.2.3(4), both bounds inclusive, the same assignment of the three printed terms to lower bound, value under check and upper bound, and the same boolean for all seven test input sets. It confirmed the check reading over the assignment reading and quoted the sentence that decides it. Two points it raised are both addressed above: it suggested the class name should say cotangent rather than angle, which is why the class carries that name, and it suggested rejecting `cot(theta_min) < 1`, which is the open decision listed above.

Note that the verification ran against the earlier `DoubleComparisonFormula` version. The arithmetic and the boundaries are unchanged by the move to `AggregatedComparisonFormula`; only the base class and the way the two bounds are exposed changed, and the test suite for both was extended rather than replaced.

## Formulas as implemented

**Formula (8.41)**, `Form8Dot41CheckCotangentCompressionFieldAngle`

`equation`

$$
CHECK \to 1 \leq \cot(\theta) \leq \cot(\theta_{min})
$$

`complete`, with `cot_theta = 2,0` and `cot_theta_min = 2,5`

$$
CHECK \to 1 \leq \cot(\theta) \leq \cot(\theta_{min}) \to 1 \leq 2.000 \leq 2.500 \to OK
$$

`complete_with_units`

$$
CHECK \to 1 \leq \cot(\theta) \leq \cot(\theta_{min}) \to 1 \leq 2.000 \leq 2.500 \to OK
$$

Every term here is dimensionless, so `complete_with_units` is deliberately the same string as `complete` rather than being left empty.

And the failing case, with `cot_theta = 3,0` against the same upper bound:

$$
CHECK \to 1 \leq \cot(\theta) \leq \cot(\theta_{min}) \to 1 \leq 3.000 \leq 2.500 \to \text{Not OK}
$$

### DIFF
--- a/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_41.py
+++ b/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_41.py
@@ -1,0 +1,193 @@
+"""Formula 8.41 from FprEN 1992-1-1:2023: Chapter 8: Ultimate limit states (ULS)."""
+
+import operator
+from collections.abc import Callable, Sequence
+from typing import Self
+
+from blueprints.codes.eurocode.fpr_en_1992_1_1_2023 import FPR_EN_1992_1_1_2023
+from blueprints.codes.formula import AggregatedComparisonFormula, ComparisonFormula
+from blueprints.codes.latex_formula import LatexFormula, latex_replace_symbols
+from blueprints.type_alias import DIMENSIONLESS
+
+
+class SubForm8Dot41LowerBound(ComparisonFormula):
+    r"""Class representing the lower bound of formula 8.41, [$1 \leq \cot\theta$]."""
+
+    label = "8.41"
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(self, cot_theta: DIMENSIONLESS) -> None:
+        r"""Check the selected inclination of the compression field against its lower bound.
+
+        FprEN 1992-1-1:2023 (E) art 8.2.3 (4) - Formula (8.41)
+
+        Parameters
+        ----------
+        cot_theta : DIMENSIONLESS
+            [$\cot\theta$] Cotangent of the selected inclination of the compression field [$-$].
+        """
+        super().__init__()
+        self.cot_theta = cot_theta
+
+    @classmethod
+    def _comparison_operator(cls) -> Callable[[float, float], bool]:
+        return operator.le
+
+    @staticmethod
+    def _evaluate_lhs(*_args, **_kwargs) -> float:
+        """Evaluates the lower bound, which the standard prints as the constant 1."""
+        return 1.0
+
+    @staticmethod
+    def _evaluate_rhs(cot_theta: DIMENSIONLESS, *_args, **_kwargs) -> float:
+        """Evaluates the value under check, for more information see the __init__ method."""
+        return float(cot_theta)
+
+    def latex(self, n: int = 3) -> LatexFormula:
+        """Returns LatexFormula object for the lower bound of formula 8.41."""
+        _equation: str = r"1 \leq \cot(\theta)"
+        return LatexFormula(
+            return_symbol=r"CHECK",
+            result="OK" if self.__bool__() else r"\text{Not OK}",
+            equation=_equation,
+            numeric_equation=latex_replace_symbols(
+                template=_equation,
+                replacements={r"\cot(\theta)": f"{self.cot_theta:.{n}f}"},
+                unique_symbol_check=False,
+            ),
+            comparison_operator_label=r"\to",
+            unit="",
+        )
+
+
+class SubForm8Dot41UpperBound(ComparisonFormula):
+    r"""Class representing the upper bound of formula 8.41, [$\cot\theta \leq \cot\theta_{min}$]."""
+
+    label = "8.41"
+    source_document = FPR_EN_1992_1_1_2023
+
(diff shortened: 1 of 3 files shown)


egarciamendez:
Closing in favour of #1095, which implements (8.41) together with (8.42) to (8.53). Two implementations of the same formula on the same path would have collided on merge, and #1095 is the better of the two.

Run side by side, they agree on every case tested: `cot_theta` of 2,0 / 1,0 / 2,5 / 0,8 / 3,0 against `cot_theta_min` = 2,5, and 2,8 against 3,0. The differences are in the API and in the validation, and #1095 wins on both counts that matter.

- **It rejects a non-physical inclination and this one does not.** This implementation reports OK for `cot_theta = -1.0`, where `1 <= -1.0 <= 2.5` is false, and raises `ZeroDivisionError` for 0.0. The cause is that the boolean of a comparison formula was derived from the ratio of its two sides, which only orders them while both have the same sign. #1095 uses `DoubleComparisonFormula`, which applies its operators directly, and `cot()` rejects angles outside 0 to 90 degrees on top of that. The root cause is fixed separately in #1105.
- **It fixes the numpy boolean at the helper rather than one formula at a time.** #1095 wraps the trigonometric helpers in `math_helpers.py` in `float()`. Without that, `DoubleComparisonFormula.__bool__` returns a `numpy.bool` and `bool(formula)` raises `TypeError`. Doing it at the helper covers every formula that uses `cot`, `sec` or `csc`.

One observation from this branch worth keeping in mind for #1095, not a blocker. 8.2.3(4) gives `cot(theta_min)` as a number: 2,5, or 3,0, or `2,5 - 0,1 * N_Ed / |V_Ed|`. A signature that takes an angle asks the caller for a conversion the standard does not, and the docstring here rounds it: 2,5 becomes 21,8 degrees, which is `cot` = 2,5002, and 3,0 becomes 18,4 degrees, which is `cot` = 3,0061. Marginally more permissive than the printed value. Every downstream formula from (8.42) to (8.62) needs `cot(theta)` rather than `theta`, so the conversion happens on the way in and again on the way out.

The blind verification behind this branch stands and applies to (8.41) either way: clause 8.2.3(4), both bounds inclusive, the constant 1 as the lower bound, and the check reading rather than an assignment, since 8.2.3(4) says the inclination "may be selected within the following range".

The one thing that depended on this branch, a docstring reference in (8.55), has been generalised to the formula number in #1098.

PR #1098, opened 2026-07-27
feat: add formulas 8.54 and 8.55 from FprEN 1992-1-1:2023 clause 8.2.3

egarciamendez:
Implements formulas (8.54) and (8.55) of FprEN 1992-1-1:2023 (E), clause 8.2.3, paragraphs (10) and (12), both on page 125.

The two are unrelated in subject, (8.54) is about ducts in the web and (8.55) about concentrated loads close to a support, but they sit on the same page of the standard, so a reviewer can check both against one page.

Contributes to #1083.

## Conditions of application are not turned into bounds

Both formulas carry a sentence that says when they apply, and neither of those sentences is a bound on the result. Turning one into a clamp would fabricate numbers the standard does not give, so neither is enforced and both have a test that locks that in.

- (8.54) applies "[w]here the web contains ducts of diameters such that `sum(phi_duct) > b_w/8`". For a smaller sum the class still returns a value; the caller decides whether the clause applies. The standard also prints no lower bound on the result, so ducts wider than the web give a negative nominal width, returned unchanged. `test_result_is_not_clamped_at_zero` pins that down.
- (8.55) applies where a concentrated load sits at a distance `a_v = z * cot(beta_incl)` less than `z * cot(theta)` from a support, which means `cot(beta_incl) < cot(theta)`. A larger `cot(beta_incl)` makes the first term negative and the result is returned as it comes out. `test_no_clamp_outside_the_condition_of_application` pins that down.

The distinction matters in practice: turning a condition of application into a floor is exactly the defect found in PR #1009 on the Dutch national annex, where `alpha > 575` became a clamp and made the value 1,00 unreachable.

## Decisions

- (8.55) is a plain `Formula` returning MPa, not a `ComparisonFormula`. The standard writes `tau_Rd = expression <= bound`, an assignment with an upper bound, and 8.2.3(12) introduces it with "the shear stress resistance may be enhanced according to". That is a value to compute, so the implementation returns `min(expression, bound)`. Compare (8.41) in #1096, where nothing is assigned and the base class is `DoubleComparisonFormula`.
- `k_duct` is a plain input. The standard gives 0,5 for grouted steel ducts, 0,8 for grouted plastic ducts with a thin wall and 1,2 for the remaining cases, with the wall thickness measured against `max{0,035 * phi_duct; 2 mm}`. None of the three carries a formula number, and the plastic case cannot be decided from a category alone since it needs the wall thickness and the duct diameter. Modelling it as a `Literal` category the way `Form5Dot1CriteriumDisregardSecondOrderEffects` does would therefore either hide that criterion or invent an API the standard does not define. All three values are in the docstring and all three have a test.
- The LaTeX of (8.55) writes `\min(a, b)` where the standard prints `a <= b`, following (8.27) and (8.32) in the earlier pull requests. It also writes `\left(\cot(\theta)\right)^2` where the standard prints `cot^2(theta)`. Those are the same quantity, and the bracketed form keeps the exponent attached to the value after substitution rather than to a unit.

## Verification

Both formulas were blindly verified against the rendered pages 122 to 125 of the standard, one agent per formula, neither of which saw the implementation or the tests.

**(8.54): verdict OK.** It independently derived clause 8.2.3(10), the same expression, and the same result for all three values of `k_duct`. It confirmed that the `sum(phi_duct) > b_w/8` sentence is a condition of application and not a bound, and that the condition uses the gross `b_w` rather than `b_w,nom`, so there is no circularity. It flagged that the class does not enforce `k_duct` being one of the three printed values, which is the documentation duty accepted above. It also noted that the standard itself mixes the glyphs for the duct diameter between the condition and the formula, which is a typesetting artefact of the standard.

**(8.55): verdict OK.** It independently derived clause 8.2.3(12), the same expression with the same fraction scope, the exponent on the cotangent rather than on the angle, and the same result for both input sets. It confirmed the bound is a cap on a computed value rather than a pass or fail check, quoting the sentence that decides it. It observed that the two input sets did not exercise `cot(beta_incl) > cot(theta)`, which is where an unwanted clamp would hide; that test was added in response. It could not confirm the graphical definition of `beta_incl` from Figure 8.11, which sits on page 126 and was not supplied, so the docstring describes `beta_incl` only through the relation the printed text gives it.

## Formulas as implemented

**Formula (8.54)**, `Form8Dot54NominalWebWidth`

`equation`

$$
b_{w,nom} = b_w - k_{duct} \cdot \Sigma\phi_{duct}
$$

`complete`

$$
b_{w,nom} = b_w - k_{duct} \cdot \Sigma\phi_{duct} = 400.000 - 0.800 \cdot 100.000 = 320.000 \ mm
$$

`complete_with_units`

$$
b_{w,nom} = b_w - k_{duct} \cdot \Sigma\phi_{duct} = 400.000 \ mm - 0.800 \cdot 100.000 \ mm = 320.000 \ mm
$$

**Formula (8.55)**, `Form8Dot55EnhancedShearStressResistance`

`equation`

$$
\tau_{Rd} = \min\left(\nu \cdot f_{cd} \cdot \frac{\cot(\theta) - \cot(\beta_{incl})}{1 + \left(\cot(\theta)\right)^2} + \rho_w \cdot f_{ywd} \cdot \cot(\beta_{incl}), \nu \cdot f_{cd} \cdot \frac{\cot(\theta)}{1 + \left(\cot(\theta)\right)^2}\right)
$$

`complete`

$$
\tau_{Rd} = \min\left(\nu \cdot f_{cd} \cdot \frac{\cot(\theta) - \cot(\beta_{incl})}{1 + \left(\cot(\theta)\right)^2} + \rho_w \cdot f_{ywd} \cdot \cot(\beta_{incl}), \nu \cdot f_{cd} \cdot \frac{\cot(\theta)}{1 + \left(\cot(\theta)\right)^2}\right) = \min\left(0.500 \cdot 20.000 \cdot \frac{2.000 - 1.000}{1 + \left(2.000\right)^2} + 0.002 \cdot 435.000 \cdot 1.000, 0.500 \cdot 20.000 \cdot \frac{2.000}{1 + \left(2.000\right)^2}\right) = 2.870 \ MPa
$$

`complete_with_units`

$$
\tau_{Rd} = \min\left(\nu \cdot f_{cd} \cdot \frac{\cot(\theta) - \cot(\beta_{incl})}{1 + \left(\cot(\theta)\right)^2} + \rho_w \cdot f_{ywd} \cdot \cot(\beta_{incl}), \nu \cdot f_{cd} \cdot \frac{\cot(\theta)}{1 + \left(\cot(\theta)\right)^2}\right) = \min\left(0.500 \cdot 20.000 \ MPa \cdot \frac{2.000 - 1.000}{1 + \left(2.000\right)^2} + 0.002 \cdot 435.000 \ MPa \cdot 1.000, 0.500 \cdot 20.000 \ MPa \cdot \frac{2.000}{1 + \left(2.000\right)^2}\right) = 2.870 \ MPa
$$

## Summary by CodeRabbit

* **New Features**
  * Added Eurocode 2:2023 calculations for nominal web width (Formula 8.54).
  * Added enhanced shear stress resistance calculations for concentrated loads near supports (Formula 8.55).
  * Added symbolic, numeric, and unit-aware formula representations.

* **Documentation**
  * Marked Formulas 8.54 and 8.55 as implemented in the Eurocode formula guide.

* **Tests**
  * Added coverage for calculations, input validation, edge cases, and formula representations.

### DIFF
--- a/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_55.py
+++ b/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_55.py
@@ -1,0 +1,118 @@
+"""Formula 8.55 from FprEN 1992-1-1:2023: Chapter 8: Ultimate limit states (ULS)."""
+
+from blueprints.codes.eurocode.fpr_en_1992_1_1_2023 import FPR_EN_1992_1_1_2023
+from blueprints.codes.formula import Formula
+from blueprints.codes.latex_formula import LatexFormula, latex_replace_symbols
+from blueprints.type_alias import DIMENSIONLESS, MPA
+from blueprints.validations import raise_if_negative
+
+
+class Form8Dot55EnhancedShearStressResistance(Formula):
+    """Class representing formula 8.55 for the calculation of the enhanced shear stress resistance where a
+    concentrated load is applied close to a support.
+    """
+
+    label = "8.55"
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(
+        self,
+        nu: DIMENSIONLESS,
+        f_cd: MPA,
+        cot_theta: DIMENSIONLESS,
+        cot_beta_incl: DIMENSIONLESS,
+        rho_w: DIMENSIONLESS,
+        f_ywd: MPA,
+    ) -> None:
+        r"""[$\tau_{Rd}$] Enhanced shear stress resistance for concentrated loads applied close to a support [$MPa$].
+
+        FprEN 1992-1-1:2023 (E) art 8.2.3 (12) - Formula (8.55)
+
+        The standard offers this enhancement in cases where concentrated loads are applied at a distance
+        [$a_v = z \cdot \cot\beta_{incl}$] less than [$z \cdot \cot\theta$] from a support. That is a condition
+        of application, not a bound on the result, so it is not enforced here.
+
+        Parameters
+        ----------
+        nu : DIMENSIONLESS
+            [$\nu$] Strength reduction factor for concrete cracked in shear. A value of 0,5 may be adopted when
+            using the angles of the compression field given in 8.2.3(4), see 8.2.3(6). A higher value may be
+            adopted under the conditions of 8.2.3(7), calculated according to Formula (8.45) [$-$].
+        f_cd : MPA
+            [$f_{cd}$] Design value of the compressive strength of concrete [$MPa$].
+        cot_theta : DIMENSIONLESS
+            [$\cot\theta$] Cotangent of the inclination of the compression field in the web, selected within the
+            range of Formula (8.41) [$-$].
+        cot_beta_incl : DIMENSIONLESS
+            [$\cot\beta_{incl}$] Cotangent of the angle [$\beta_{incl}$] that follows from the distance of the
+            concentrated load to the support through [$a_v = z \cdot \cot\beta_{incl}$] [$-$].
+        rho_w : DIMENSIONLESS
+            [$\rho_w$] Shear reinforcement ratio according to Formula (8.43) [$-$].
+        f_ywd : MPA
+            [$f_{ywd}$] Design value of the yield strength of the shear reinforcement [$MPa$].
+        """
+        super().__init__()
+        self.nu = nu
+        self.f_cd = f_cd
+        self.cot_theta = cot_theta
+        self.cot_beta_incl = cot_beta_incl
+        self.rho_w = rho_w
+        self.f_ywd = f_ywd
+
+    @staticmethod
+    def _evaluate(
+        nu: DIMENSIONLESS,
+        f_cd: MPA,
+        cot_theta: DIMENSIONLESS,
+        cot_beta_incl: DIMENSIONLESS,
+        rho_w: DIMENSIONLESS,
+        f_ywd: MPA,
+    ) -> MPA:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(nu=nu, f_cd=f_cd, cot_theta=cot_theta, cot_beta_incl=cot_beta_incl, rho_w=rho_w, f_ywd=f_ywd)
+
+        # The denominator 1 + cot^2(theta) is never zero, so it needs no guard of its own.
+        enhanced = nu * f_cd * (cot_theta - cot_beta_incl) / (1 + cot_theta**2) + rho_w * f_ywd * cot_beta_incl
+        limit = nu * f_cd * cot_theta / (1 + cot_theta**2)
+        return min(enhanced, limit)
+
+    def latex(self, n: int = 3) -> LatexFormula:
+        """Returns LatexFormula object for formula 8.55."""
+        _equation: str = (
+            r"\min\left(\nu \cdot f_{cd} \cdot \frac{\cot(\theta) - \cot(\beta_{incl})}"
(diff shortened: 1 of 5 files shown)


tests/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/test_formula_8_54.py
```diff
@@ -0,0 +1,90 @@
...
+        # Example values, with the ducts wider than the web itself
+        formula = Form8Dot54NominalWebWidth(b_w=400.0, k_duct=1.2, sum_phi_duct=400.0)
+
+        # 400 - 1.2 * 400 = -80
+        assert formula == pytest.approx(expected=-80.0, rel=1e-4)
+
+    @pytest.mark.parametrize(
+        ("b_w", "k_duct", "sum_phi_duct"),
+        [
+            (400.0, -0.8, 100.0),  # k_duct is negative
```
casper-van-der-krift: Given that k_duct is restricted to be either 0.5, 0.8 or 1.2, I would think that k_duct cannot be zero. So k_duct should be tested in `test_raise_error_when_less_or_equal_to_zero` instead of `test_raise_error_when_negative_values_are_given`.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_55.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+
+    label = "8.55"
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(
+        self,
+        nu: DIMENSIONLESS,
+        f_cd: MPA,
+        cot_theta: DIMENSIONLESS,
```
casper-van-der-krift: Shouldn't we use theta and beta_incl as variables here and take cot in the formula itself?

casper-van-der-krift (changes requested):

tests/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/test_formula_8_54.py
egarciamendez: Done, moved to `test_raise_error_when_less_or_equal_to_zero` together with the negative case, and the guard in the formula changed to `raise_if_less_or_equal_to_zero(b_w=b_w, k_duct=k_duct)`.

Not enforcing membership of {0,5; 0,8; 1,2} itself: a float equality check on a coefficient is brittle, and the choice between the three is a classification of the duct that belongs to the caller.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_55.py
egarciamendez: Yes, changed to `theta: DEG` and `beta_incl: DEG` with `cot()` from `math_helpers`. You are right that I was the outlier: 8.41, 8.44, 8.45 and 8.50 in this same clause all take the angle, as do nine formulas in EN 1992-1-1:2004 chapter 6.

Side effects: `cot()` already rejects angles outside (0, 90], so a zero angle no longer divides by zero; LaTeX now prints `\cot(26.565)` instead of the pre-computed `2.000`, and the units form uses `^\circ` like 8.44 does. Tests converted to angles with round cotangents.

PR #1099, opened 2026-07-27
feat: add formulas 8.56 and 8.57 from FprEN 1992-1-1:2023 clause 8.2.3

egarciamendez:
Implements formulas (8.56) and (8.57) of FprEN 1992-1-1:2023 (E), clause 8.2.3(12), both on page 126.

Contributes to #1083.

## The paragraph number was checked, not assumed

Page 126 carries no paragraph marker at all. It opens mid-sentence, "but should not be larger than `cot(theta)` according to 8.2.3(5), NOTE", which completes NOTE 1 of paragraph (12) begun at the bottom of page 125, and it then runs through (8.56), (8.57), NOTE 2 and Figure 8.11 without a new marker.

Both blind verifiers reached 8.2.3(12) and both flagged that they had inferred it. That was then checked against the paragraph markers themselves: page 125 carries (10), (11) and (12), page 126 carries none, and (13) begins on page 127. So (8.56) and (8.57) are the closing content of 8.2.3(12).

## Decisions

- Both are plain `Formula` classes. (8.56) prints `sigma_swd = expression <= f_ywd`, an assignment with an upper bound, so it returns `min(expression, f_ywd)`. (8.57) prints no bound at all.
- Neither result is clamped at zero, because the standard prints no lower bound on either.
  - (8.56) turns negative when `cot(theta)^2 * (epsilon_x + 0,001)` falls below 0,001, which happens for a small cotangent together with a small longitudinal strain. `test_result_is_not_clamped_at_zero` pins that down at -150,0 MPa.
  - (8.57) is a moment to be added to `M_Ed`, so a negative value is meaningful: it reduces the chord forces. It turns negative where the investigated cross-section lies beyond half the distance to the concentrated force, and again where the average shear stress falls below the contribution of the shear reinforcement. Both cases have their own test.
- (8.56) exists precisely for `cot(theta) < 1`, which lies outside the range that Formula (8.41) allows. That is a condition of application and not a bound, so `cot_theta` is not clamped to 1 here. Formula (8.41) is the general rule that this paragraph explicitly relaxes.
- In (8.57) the `f_ywd` slot receives the stress `sigma_swd` from (8.56) whenever `cot(theta) < 1`. The formula itself is unchanged; only the meaning of the stress changes, so the substitution is the caller's, and the docstring says so.
- The LaTeX writes `\min(a, b)` where (8.56) prints `a <= b`, and `\left(\cot(\theta)\right)^2` where the standard prints `cot^2(theta)`, both following the earlier pull requests in this series.

## Verification

Both formulas were blindly verified against the rendered pages 122 to 126 of the standard, one agent per formula, neither of which saw the implementation or the tests.

**(8.56): verdict OK.** Same expression, same result for all three input sets, including the negative one. It set out the bracket nesting independently and in full: `E_s` multiplies the whole square bracket, the first 0,001 sits inside the parenthesis and is therefore multiplied by `cot(theta)^2`, and the second 0,001 sits outside that parenthesis and is not. It confirmed the exponent is on the cotangent rather than on the angle, that the bound is a cap and not a pass or fail check, that the standard prints no lower bound, and that clamping `cot_theta` to 1 would be wrong.

**(8.57): verdict OK.** Same expression, same result for all three input sets, including both negative ones. It confirmed the bracket scope from the printed layout, in particular that the fraction bar sits under `a` alone so the division by 2 does not apply to `a - x`, and it derived the unit Nmm from the dimensions of the factors. It warned that a caller could confuse `a`, the distance from the support axis to the concentrated force, with the shear span `a_v = z * cot(beta_incl)` of (8.55); the docstring now says explicitly that they are different quantities.

## Formulas as implemented

**Formula (8.56)**, `Form8Dot56StressInShearReinforcement`

`equation`

$$
\sigma_{swd} = \min\left(E_s \cdot \left[\left(\cot(\theta)\right)^2 \cdot \left(\varepsilon_x + 0.001\right) - 0.001\right], f_{ywd}\right)
$$

`complete`

$$
\sigma_{swd} = \min\left(E_s \cdot \left[\left(\cot(\theta)\right)^2 \cdot \left(\varepsilon_x + 0.001\right) - 0.001\right], f_{ywd}\right) = \min\left(200000.000 \cdot \left[\left(0.800\right)^2 \cdot \left(0.001 + 0.001\right) - 0.001\right], 435.000\right) = 56.000 \ MPa
$$

`complete_with_units`

$$
\sigma_{swd} = \min\left(E_s \cdot \left[\left(\cot(\theta)\right)^2 \cdot \left(\varepsilon_x + 0.001\right) - 0.001\right], f_{ywd}\right) = \min\left(200000.000 \ MPa \cdot \left[\left(0.800\right)^2 \cdot \left(0.001 + 0.001\right) - 0.001\right], 435.000 \ MPa\right) = 56.000 \ MPa
$$

**Formula (8.57)**, `Form8Dot57AdditionalBendingMoment`

`equation`

$$
\Delta M_{Ed} = \left(\tau_{Ed} - \rho_w \cdot f_{ywd} \cdot \cot(\theta)\right) \cdot z \cdot b_w \cdot \left(\frac{a}{2} - x\right)
$$

`complete`

$$
\Delta M_{Ed} = \left(\tau_{Ed} - \rho_w \cdot f_{ywd} \cdot \cot(\theta)\right) \cdot z \cdot b_w \cdot \left(\frac{a}{2} - x\right) = \left(2.000 - 0.002 \cdot 435.000 \cdot 2.000\right) \cdot 400.000 \cdot 300.000 \cdot \left(\frac{600.000}{2} - 200.000\right) = 3120000.000 \ Nmm
$$

`complete_with_units`

$$
\Delta M_{Ed} = \left(\tau_{Ed} - \rho_w \cdot f_{ywd} \cdot \cot(\theta)\right) \cdot z \cdot b_w \cdot \left(\frac{a}{2} - x\right) = \left(2.000 \ MPa - 0.002 \cdot 435.000 \ MPa \cdot 2.000\right) \cdot 400.000 \ mm \cdot 300.000 \ mm \cdot \left(\frac{600.000 \ mm}{2} - 200.000 \ mm\right) = 3120000.000 \ Nmm
$$

## Summary by CodeRabbit

* **New Features**
  * Added Eurocode 2 Formula 8.56 for calculating stress in shear reinforcement.
  * Added Formula 8.57 for calculating additional bending moments in support regions.
  * Added symbolic and unit-aware formula representations.

* **Documentation**
  * Marked formulas 8.56 and 8.57 as completed in the Eurocode 2 formula guide.

* **Tests**
  * Added coverage for calculations, validation, negative results, and formula representations.


blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_56.py
```diff
@@ -0,0 +1,89 @@
...
+        Compression field inclinations with [$\cot\theta < 1$] are allowed if the yield strength [$f_{ywd}$] in
+        Formulae (8.55) and (8.57) is replaced by this stress. That is a condition of application, not a bound,
+        so it is not enforced here. The standard prints no lower bound on the result either, so a combination
+        that gives a negative stress returns that value unchanged.
+
+        Parameters
+        ----------
+        e_s : MPA
+            [$E_s$] Modulus of elasticity of the reinforcing steel [$MPa$].
+        cot_theta : DIMENSIONLESS
```
casper-van-der-krift: Same comment as on formula 8.55: the variable should be theta, instead of cot(theta).

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_57.py
```diff
@@ -0,0 +1,131 @@
...
+
+    label = "8.57"
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(
+        self,
+        tau_ed: MPA,
+        rho_w: DIMENSIONLESS,
+        f_ywd: MPA,
+        cot_theta: DIMENSIONLESS,
```
casper-van-der-krift: Same comment as in 8.55 and 8.56: variable should be theta instead of cot(theta).

casper-van-der-krift (changes requested):

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_56.py
egarciamendez: Changed to `theta: DEG` with `cot()` from `math_helpers`, same as 8.55. Guarded with `raise_if_less_or_equal_to_zero`, so the range 0 < theta <= 90 now holds here too.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_57.py
egarciamendez: Same change here: `theta: DEG` instead of `cot_theta`.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_57.py
```diff
@@ -0,0 +1,131 @@
...
+        z: MM,
+        b_w: MM,
+        a: MM,
+        x: MM,
+    ) -> NMM:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(tau_ed=tau_ed, rho_w=rho_w, f_ywd=f_ywd, cot_theta=cot_theta, z=z, b_w=b_w, a=a, x=x)
+
+        return (tau_ed - rho_w * f_ywd * cot_theta) * z * b_w * (a / 2 - x)
```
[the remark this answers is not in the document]
egarciamendez: Correct. Confirmed against the page: (8.46) prints `epsilon_x >= 0` but (8.56) prints only `<= f_ywd`, so with `cot(theta) < 1` a negative `sigma_swd` is the normal outcome, not an edge case. Dropped `f_ywd` from `raise_if_negative` and replaced the rejection test with one that feeds the negative 8.56 result in.

Note that 8.55 on main carries the same guard while documenting the same substitution. Left out of this PR.

PR #1100, opened 2026-07-27
feat: add formulas 8.58, 8.59 and 8.60 from FprEN 1992-1-1:2023 clause 8.2.3

egarciamendez:
Implements formulas (8.58), (8.59) and (8.60) of FprEN 1992-1-1:2023 (E), clause 8.2.3(13), page 127.

Paragraph (13) covers members with inclined shear reinforcement, `45 <= alpha_w < 90` degrees, and replaces four earlier formulas: (8.41), (8.42), (8.44) and (8.50) become (8.58), (8.59), (8.60) and (8.61) respectively. This pull request takes the first three; (8.61) follows.

Contributes to #1083.

## Base classes: (8.58) aggregates, (8.60) compares

`DoubleComparisonFormula` is on its way out, so the two-sided range of (8.58) is built with
`AggregatedComparisonFormula` instead. That is not a like-for-like swap. The aggregate composes several
independent `ComparisonFormula` instances with `all` or `any`, so the one printed relation
`tan(alpha_w/2) <= cot(theta) <= cot(theta_min)` becomes two of them, `SubForm8Dot58LowerBound` and
`SubForm8Dot58UpperBound`, joined with `all`.

What that buys, and what it costs:

- Each bound now carries its own `unity_check`, and the aggregate reports the larger of the two, so the
  governing bound is visible. `DoubleComparisonFormula` has no unity check at all. For `cot(theta) = 0.5` at
  `alpha_w = 60` degrees the lower bound governs at 1.155; for `cot(theta) = 3.0` the upper bound governs at 1.20.
- `lhs` and `rhs` raise `NotImplementedError` on the aggregate by design, so the two halves are reached through
  `comparison_formulas` instead. There is a test that pins down all four values.
- `__new__` has to translate the three arguments of this formula into the `aggregation` and `comparison_formulas`
  pair the base class evaluates. Overriding `_evaluate` instead does not work: its signature is fixed by the base
  class and `ty` rejects the narrowed override.

## Why (8.60) is a ComparisonFormula

Every numbered formula in this series so far has been a value to compute. (8.60) is not: it is a verification that passes or fails, and it is the first plain `ComparisonFormula` here.

The paragraph itself only says that (8.60) replaces (8.44), so the meaning comes from the sentence that introduces (8.44) in 8.2.3(5): "The stress in the compression field in all cross-sections shall be verified according to:". Shall be verified, not may be taken as. The class therefore returns a boolean, exposes the compression field stress as `lhs`, its limit as `rhs`, and their ratio as `unity_check`.

That is the distinction this series has been applying throughout, now with three different base classes in use side by side:

| Formula | Printed as | Base class | Why |
|---|---|---|---|
| (8.58) | `tan(alpha_w/2) <= cot(theta) <= cot(theta_min)` | `AggregatedComparisonFormula` | a range the designer's choice must lie in, nothing is assigned |
| (8.59) | `tau_Rd,sy = expression` | `Formula` | a value, no bound |
| (8.60) | `sigma_cd = expression <= nu * f_cd` | `ComparisonFormula` | an action against a resistance, introduced with "shall be verified" |

Compare (8.55) and (8.62), where the same `<=` genuinely caps a computed resistance because the lead-in reads "the shear stress resistance may be enhanced according to". The operator does not decide it; the sentence above the formula does.

## Decisions

- `alpha_w` is passed as the angle rather than as its cotangent. Both `cot(alpha_w)` and `sin(alpha_w)` appear in (8.59), and passing two derived values would let a caller supply a pair that cannot come from one angle.
- `cot(theta)` stays a cotangent, because that is what the standard treats as the free variable, bounded by (8.58). So the two angles enter differently on purpose, and the docstrings say so.
- The printed range `45 <= alpha_w < 90` degrees is a condition of application, not a bound, so it is not enforced. Three tests do use `alpha_w = 90` and are labelled as transcription checks rather than supported use: there `tan(45) = 1`, `cot(90) = 0` and `sin(90) = 1`, so each formula collapses onto the one it replaces. That is the strongest available evidence that the transcription is right, and the blind verifier of (8.60) independently used the same check.
- (8.60) guards its denominator `cot(theta) + cot(alpha_w)`. Within the printed range it is strictly positive, but above 90 degrees the cotangent turns negative and the expression would otherwise return a negative stress with no sign that the input was out of range. The standard says such angles should be avoided, so raising is the honest response rather than reporting a passing check.

## Two defects the tests caught

- `ComparisonFormula.__bool__` rejects a numpy boolean. Because `_evaluate_lhs` used `np.tan`, the unity check came out as a numpy type and Python raised `TypeError: __bool__ should return bool, returned numpy.bool`. Both sides are now cast to `float`. Any comparison formula that reaches for numpy will hit this.
- The first version of the denominator test used `cot_theta = 0` with `alpha_w = 90`, expecting the denominator to vanish. It does not: `tan(90 degrees)` is finite in floating point, around `1.6e16`, so its reciprocal is `6e-17` rather than zero. The test now uses `alpha_w = 120`, where the cotangent is genuinely negative, which is the case the guard exists for.

## Verification

All three were blindly verified against the rendered pages 122 to 127 of the standard, one agent per formula, none of which saw the implementation or the tests.

**(8.58): verdict OK.** Same seven boolean results, same numeric lower bound for each of the three angles used. It confirmed both bounds are inclusive, and confirmed independently that the angle is halved before the tangent is taken rather than after, citing the text under (8.62) as corroboration. It noted that `tan(radians(45))` returns `0.9999999999999999`, so the lower bound at 90 degrees sits marginally below 1; the error happens to fall on the safe side for an inclusive bound.

**(8.59): verdict OK.** Same result for both input sets, with its arithmetic shown. It confirmed the sine multiplies the whole product and sits outside the bracket. It pointed out that the 90 degree case cannot catch a misplaced `cot(alpha_w)` because that term vanishes there, so the 60 degree case is the discriminating one; both are present, and the test comment now says which is which.

**(8.60): verdict OK.** Same left-hand side and same verdict for both input sets. On the central question it reached the verification reading independently, quoted the deciding sentence from 8.2.3(5), and set it against (8.55), (8.62) and (8.45) where the same operator means a cap instead. It confirmed the numerator is `1 + cot(theta)^2`, the denominator `cot(theta) + cot(alpha_w)`, and the exponent sits on the cotangent. It also warned that a `ComparisonFormula` returning only a boolean would lose the stress, which is why `lhs` carries the full expression rather than just `tau_Ed`.

## Formulas as implemented

**Formula (8.58)**, `Form8Dot58CheckCotangentInclinedShearReinforcement`

`equation`

$$
CHECK \to \tan\left(\frac{\alpha_w}{2}\right) \leq \cot(\theta) \leq \cot(\theta_{min})
$$

`complete`

$$
CHECK \to \tan\left(\frac{\alpha_w}{2}\right) \leq \cot(\theta) \leq \cot(\theta_{min}) \to \tan\left(\frac{60.000}{2}\right) \leq 2.000 \leq 2.500 \to OK
$$

`complete_with_units`

$$
CHECK \to \tan\left(\frac{\alpha_w}{2}\right) \leq \cot(\theta) \leq \cot(\theta_{min}) \to \tan\left(\frac{60.000 \ degrees}{2}\right) \leq 2.000 \leq 2.500 \to OK
$$

**Formula (8.59)**, `Form8Dot59ShearStressResistanceInclinedShearReinforcement`

`equation`

$$
\tau_{Rd,sy} = \rho_w \cdot f_{ywd} \cdot \left(\cot(\theta) + \cot(\alpha_w)\right) \cdot \sin(\alpha_w)
$$

`complete`

$$
\tau_{Rd,sy} = \rho_w \cdot f_{ywd} \cdot \left(\cot(\theta) + \cot(\alpha_w)\right) \cdot \sin(\alpha_w) = 0.002 \cdot 435.000 \cdot \left(2.000 + \cot(60.000)\right) \cdot \sin(60.000) = 1.942 \ MPa
$$

`complete_with_units`

$$
\tau_{Rd,sy} = \rho_w \cdot f_{ywd} \cdot \left(\cot(\theta) + \cot(\alpha_w)\right) \cdot \sin(\alpha_w) = 0.002 \cdot 435.000 \ MPa \cdot \left(2.000 + \cot(60.000 \ degrees)\right) \cdot \sin(60.000 \ degrees) = 1.942 \ MPa
$$

**Formula (8.60)**, `Form8Dot60CheckCompressionFieldStressInclinedShearReinforcement`

`equation`

$$
CHECK \to \sigma_{cd} = \tau_{Ed} \cdot \frac{1 + \left(\cot(\theta)\right)^2}{\cot(\theta) + \cot(\alpha_w)} \leq \nu \cdot f_{cd}
$$

`complete`

$$
CHECK \to \sigma_{cd} = \tau_{Ed} \cdot \frac{1 + \left(\cot(\theta)\right)^2}{\cot(\theta) + \cot(\alpha_w)} \leq \nu \cdot f_{cd} \to 3.880 = 2.000 \cdot \frac{1 + \left(2.000\right)^2}{2.000 + \cot(60.000)} \leq 0.500 \cdot 20.000 \to OK
$$

`complete_with_units`

$$
CHECK \to \sigma_{cd} = \tau_{Ed} \cdot \frac{1 + \left(\cot(\theta)\right)^2}{\cot(\theta) + \cot(\alpha_w)} \leq \nu \cdot f_{cd} \to 3.880 \ MPa = 2.000 \ MPa \cdot \frac{1 + \left(2.000\right)^2}{2.000 + \cot(60.000 \ degrees)} \leq 0.500 \cdot 20.000 \ MPa \to OK
$$

## Summary by CodeRabbit

* **New Features**
  * Added Eurocode formulas 8.58–8.60 for inclined shear reinforcement.
  * Supports checks for compression-field inclination, shear stress resistance, and compression-field stress.
  * Added validation and LaTeX output with numeric values and units.

* **Documentation**
  * Marked formulas 8.58–8.60 as implemented and linked them to their corresponding calculations.

* **Tests**
  * Added comprehensive coverage for calculations, validation errors, boundary conditions, and formatted output.


blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
```diff
@@ -0,0 +1,200 @@
...
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(self, cot_theta: DIMENSIONLESS, alpha_w: DEG) -> None:
+        r"""Check the selected inclination of the compression field against its lower bound.
+
+        FprEN 1992-1-1:2023 (E) art 8.2.3 (13) - Formula (8.58)
+
+        Parameters
+        ----------
+        cot_theta : DIMENSIONLESS
```
casper-van-der-krift: SEC about passing angles directly, not as cotangens.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
```diff
@@ -0,0 +1,200 @@
...
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(self, cot_theta: DIMENSIONLESS, cot_theta_min: DIMENSIONLESS) -> None:
+        r"""Check the selected inclination of the compression field against its upper bound.
+
+        FprEN 1992-1-1:2023 (E) art 8.2.3 (13) - Formula (8.58)
+
+        Parameters
+        ----------
+        cot_theta : DIMENSIONLESS
```
casper-van-der-krift: SEC about passing angles directly, not as cotangens.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
```diff
@@ -0,0 +1,200 @@
...
+            [$\cot\theta$] Cotangent of the selected inclination of the compression field [$-$].
+        cot_theta_min : DIMENSIONLESS
```
casper-van-der-krift: SEC about passing angles directly, not as cotangens.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
```diff
@@ -0,0 +1,200 @@
...
+        members with inclined shear reinforcement.
+
+        FprEN 1992-1-1:2023 (E) art 8.2.3 (13) - Formula (8.58)
+
+        The standard lets the designer select the inclination, so this is a verification of that choice and not
+        a value to be calculated. Both bounds are inclusive as printed.
+
+        Parameters
+        ----------
+        cot_theta : DIMENSIONLESS
```
casper-van-der-krift: SEC about passing angles directly, not as cotangens.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
```diff
@@ -0,0 +1,200 @@
...
+            [$\cot\theta$] Cotangent of the selected inclination of the compression field in the web
+            carrying shear [$-$].
+        cot_theta_min : DIMENSIONLESS
```
casper-van-der-krift: SEC about passing angles directly, not as cotangens.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_59.py
```diff
@@ -0,0 +1,91 @@
...
+
+        FprEN 1992-1-1:2023 (E) art 8.2.3 (13) - Formula (8.59)
+
+        Parameters
+        ----------
+        rho_w : DIMENSIONLESS
+            [$\rho_w$] Shear reinforcement ratio according to Formula (8.43) [$-$].
+        f_ywd : MPA
+            [$f_{ywd}$] Design value of the yield strength of the shear reinforcement [$MPa$].
+        cot_theta : DIMENSIONLESS
```
casper-van-der-krift: SEC about passing angles directly, not as cotangens.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_59.py
```diff
@@ -0,0 +1,91 @@
...
+        self.alpha_w = alpha_w
+
+    @staticmethod
+    def _evaluate(rho_w: DIMENSIONLESS, f_ywd: MPA, cot_theta: DIMENSIONLESS, alpha_w: DEG) -> MPA:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(rho_w=rho_w, f_ywd=f_ywd, cot_theta=cot_theta)
+        raise_if_less_or_equal_to_zero(alpha_w=alpha_w)
+
+        alpha_w_rad = np.deg2rad(alpha_w)
+        return rho_w * f_ywd * (cot_theta + 1 / np.tan(alpha_w_rad)) * np.sin(alpha_w_rad)
```
casper-van-der-krift: Better to stay as close to the formula as possible by using `utils.math_helpers.cot`

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_60.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(self, tau_ed: MPA, cot_theta: DIMENSIONLESS, alpha_w: DEG, nu: DIMENSIONLESS, f_cd: MPA) -> None:
+        r"""Verify the stress in the compression field against its limit.
+
+        FprEN 1992-1-1:2023 (E) art 8.2.3 (13) - Formula (8.60)
+
+        Parameters
+        ----------
+        tau_ed : MPA
+            [$\tau_{Ed}$] Average shear stress over the cross-section according to Formula (8.18) [$MPa$].
+        cot_theta : DIMENSIONLESS
```
casper-van-der-krift: SEC about passing angles directly, not as cotangens.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_60.py
```diff
@@ -0,0 +1,121 @@
...
+    @staticmethod
+    def _evaluate_lhs(tau_ed: MPA, cot_theta: DIMENSIONLESS, alpha_w: DEG, *_args, **_kwargs) -> float:
+        """Evaluates the stress in the compression field, for more information see the __init__ method."""
+        raise_if_negative(tau_ed=tau_ed, cot_theta=cot_theta)
+        raise_if_less_or_equal_to_zero(alpha_w=alpha_w)
+
+        # Negative for the angles above 90 degrees that the standard says should be avoided, where the
+        # expression would otherwise return a negative stress without any sign that the input was outside
+        # the range this clause covers.
+        denominator = cot_theta + 1 / np.tan(np.deg2rad(alpha_w))
```
casper-van-der-krift: Better to stay as close to the formula as possible by using `utils.math_helpers.cot`

casper-van-der-krift (changes requested):

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
egarciamendez: Changed to `theta: DEG` with `cot()` from `math_helpers`, same as 8.55. All three formulas of the clause now take angles.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
egarciamendez: Done, `theta: DEG` here too.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
egarciamendez: Done, `theta_min: DEG`.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
egarciamendez: Done, `theta: DEG`.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
egarciamendez: Done, `theta_min: DEG`.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_59.py
egarciamendez: Done, `theta: DEG`.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_59.py
egarciamendez: Now `cot(theta) + cot(alpha_w)` via `math_helpers.cot`. Note it rejects angles above 90 degrees, which the standard says should be avoided.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_60.py
egarciamendez: Done, `theta: DEG`.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_60.py
egarciamendez: Same here, the denominator is now `cot(theta) + cot(alpha_w)`. That also removed the explicit denominator guard: with both angles in (0, 90] the sum cannot reach zero.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @staticmethod
+    def _evaluate_lhs(alpha_w: DEG, *_args, **_kwargs) -> float:
+        """Evaluates the lower bound, for more information see the __init__ method."""
+        return float(np.tan(np.deg2rad(alpha_w) / 2))
+
+    @staticmethod
+    def _evaluate_rhs(cot_theta: DIMENSIONLESS, *_args, **_kwargs) -> float:
+        """Evaluates the value under check, for more information see the __init__ method."""
+        return float(cot_theta)
```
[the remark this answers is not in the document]
egarciamendez: Fixed, but through the type rather than an extra guard: the parameters are now `theta: DEG` and `alpha_w: DEG`, guarded to 0 < angle <= 90. A non-positive cotangent can no longer be passed in, so the ratio has no way to go negative or divide by zero.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_60.py
```diff
@@ -0,0 +1,121 @@
...
+        # Cast away the numpy type: ComparisonFormula.__bool__ returns the result of the unity check, and
+        # Python rejects a numpy.bool there.
+        return float(tau_ed * (1 + cot_theta**2) / denominator)
+
+    @staticmethod
+    def _evaluate_rhs(nu: DIMENSIONLESS, f_cd: MPA, *_args, **_kwargs) -> float:
+        """Evaluates the limit on the stress, for more information see the __init__ method."""
+        raise_if_negative(nu=nu, f_cd=f_cd)
+
+        return float(nu * f_cd)
```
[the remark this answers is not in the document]
egarciamendez: Correct, the limit is the denominator of the unity check. `raise_if_negative` replaced by `raise_if_less_or_equal_to_zero(nu=nu, f_cd=f_cd)`, with tests for both at zero.

tests/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/test_formula_8_58.py
```diff
@@ -0,0 +1,160 @@
...
+        ],
+    )
+    def test_evaluation(self, cot_theta: float, cot_theta_min: float, alpha_w: float, expected: bool) -> None:
+        """Tests the evaluation of the result."""
+        # Create object to test
+        formula = Form8Dot58CheckCotangentInclinedShearReinforcement(cot_theta=cot_theta, cot_theta_min=cot_theta_min, alpha_w=alpha_w)
+
+        # Perform test by assert
+        assert bool(formula) is expected
+
```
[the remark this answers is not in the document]
egarciamendez: Added: six cases for angles at or below zero and three for angles above 90 degrees, on the aggregate so both sub-formulas are covered.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_58.py
```diff
@@ -0,0 +1,212 @@
...
+    def _comparison_operator(cls) -> Callable[[float, float], bool]:
+        return operator.le
+
+    @staticmethod
+    def _evaluate_lhs(alpha_w: DEG, *_args, **_kwargs) -> float:
+        """Evaluates the lower bound, for more information see the __init__ method."""
+        raise_if_less_or_equal_to_zero(alpha_w=alpha_w)
+        raise_if_greater_than_90(alpha_w=alpha_w)
+
+        return float(np.tan(np.deg2rad(alpha_w) / 2))
```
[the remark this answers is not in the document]
egarciamendez: Not applying this.

`45 <= alpha_w < 90` is the scope of clause 8.2.3(13), which names the members it addresses, not a bound printed on (8.58), (8.59) or (8.60). This repo documents such conditions and does not raise on them: 8.29, 8.54, 8.55, 8.56 and 8.76 all say so in their docstrings and all are on main.

`alpha_w = 90` is kept deliberately. There the three formulas degenerate into the (8.41), (8.42) and (8.44) they replace, because tan(45) = 1 and cot(90) = 0. That is the transcription check the tests rely on, and it is the only independent evidence that the formulas were typed correctly. Turning it into a rejection test removes the check and replaces it with nothing.

What is enforced is `alpha_w > 90`, since the standard says those angles should be avoided and the cotangent turns negative there, which would silently corrupt the result rather than merely fall outside scope.

PR #1101, opened 2026-07-27
feat: add formulas 8.61, 8.62 and 8.63 from FprEN 1992-1-1:2023 clause 8.2.3

egarciamendez:
Implements formulas (8.61), (8.62) and (8.63) of FprEN 1992-1-1:2023 (E), clause 8.2.3(13), page 127.

With this, paragraph (13) is complete. It covers members with inclined shear reinforcement, `45 <= alpha_w < 90` degrees, and every formula in it is now implemented: (8.58), (8.59) and (8.60) in #1100, and (8.61), (8.62) and (8.63) here. Together they replace (8.50), (8.55) and (8.56) for such members.

Contributes to #1083.

## The trap on this page

(8.63) is (8.56) with one factor changed, and the change is easy to transcribe wrongly. Where (8.56) has `cot(theta)^2`, (8.63) has `(cot(theta) + cot(alpha_w))^2 / (1 + cot(alpha_w)^2)`. Everything else is identical: the same `E_s`, the same `(epsilon_x + 0,001)`, the same trailing `- 0,001` inside the bracket, the same `<= f_ywd`.

The denominator is `1 + cot(alpha_w)^2`, with the angle of the reinforcement. The three formulas printed immediately around it, (8.55), (8.60) and (8.62), all carry `1 + cot(theta)^2` instead, with the angle of the compression field. The blind verifier of (8.63) rendered the formula at three times magnification to settle it and called this "the commonest transcription trap on this page". The implementation uses `cot(alpha_w)`, and at `alpha_w = 90` degrees it collapses onto (8.56) exactly, which is the check that catches the mistake.

## Decisions

- All three are plain `Formula` classes. (8.62) and (8.63) print an assignment with an upper bound, so they return `min(expression, bound)`. (8.61) prints no bound at all. Compare (8.60) in #1100, which is a `ComparisonFormula` because its lead-in reads "shall be verified".
- (8.61) takes the magnitude of the shear force as printed, so the sign of `v_ed` does not reach the result and there is a test for that.
- No result is clamped at zero. The standard prints no lower bound on any of the three.
  - (8.61) turns negative where the shear reinforcement is inclined more steeply than the compression field. The test for it uses `alpha_w = 45` degrees with `cot(theta) = 0.5`, which is **inside** the range Formula (8.58) permits, since `tan(22.5) = 0.414` lets that value through while `cot(45) = 1` still exceeds it. So a negative result is reachable in ordinary design, not only through out of range input. The verifier pointed this out: at `alpha_w = 60` degrees the two happen to coincide at `1/sqrt(3)`, which would have hidden the sign entirely.
  - (8.63) turns negative for a small strain together with a small cotangent, at -150,0 MPa in the test.
  - (8.62) turns negative in its first term where `cot(beta_incl) > cot(theta)`, which is outside the condition of application, and the result is returned as it comes out.
- `alpha_w` is passed as the angle, `cot(theta)` and `cot(beta_incl)` as cotangents. That mixed convention is deliberate and is what the standard treats as free: `alpha_w` appears through both its cotangent and its sine, while `cot(theta)` is the quantity that (8.58) bounds directly. The verifier of (8.62) flagged the mixture as easy to misuse, so every docstring says which is which.

## Verification

All three were blindly verified against the rendered pages 122 to 127 of the standard, one agent per formula, none of which saw the implementation or the tests.

**(8.61): verdict OK.** Same result for all four input sets, with its arithmetic shown. It confirmed the shear force enters as a magnitude and that the two cotangents are subtracted, contrasting that with the plus sign in (8.59), (8.60) and (8.62) on the same page. It observed that a negative result is physically meaningful because `N_Vd` feeds (8.51) and (8.52) as `N_Vd + N_Ed`, which 8.2.3(8) discusses by sign, so clamping would be an addition to the standard. Its suggestion for a better sign test was adopted, as described above. It also noted that the class name says "tensile force" while the quantity is not necessarily tensile; that wording is the standard's own, in 8.2.3(8).

**(8.62): verdict OK.** Same result for both input sets and the same value for the bound. It confirmed that the denominator of the second line is the same `1 + cot(theta)^2` as the first and that only the numerator changes, that the sine multiplies the whole bracket and not the first term, and that the exponent sits on the cotangent. It read the inequality as a cap on a computed resistance, quoting 8.2.3(12) "the shear stress resistance may be enhanced according to", and set that against (8.44) and (8.60) where the same operator means a verification.

**(8.63): verdict OK.** Same result for all three input sets, including the negative one. It set out the bracket nesting in full and confirmed the denominator carries `alpha_w` rather than `theta`, as described above. It compared (8.63) with (8.56) term by term and reported that the single changed factor is the one named above.

## Formulas as implemented

**Formula (8.61)**, `Form8Dot61AdditionalTensileForceInclinedShearReinforcement`

`equation`

$$
N_{Vd} = \left|V_{Ed}\right| \cdot \left(\cot(\theta) - \cot(\alpha_w)\right)
$$

`complete`

$$
N_{Vd} = \left|V_{Ed}\right| \cdot \left(\cot(\theta) - \cot(\alpha_w)\right) = \left|300000.000\right| \cdot \left(2.000 - \cot(60.000)\right) = 426794.919 \ N
$$

`complete_with_units`

$$
N_{Vd} = \left|V_{Ed}\right| \cdot \left(\cot(\theta) - \cot(\alpha_w)\right) = \left|300000.000 \ N\right| \cdot \left(2.000 - \cot(60.000 \ degrees)\right) = 426794.919 \ N
$$

**Formula (8.62)**, `Form8Dot62EnhancedShearStressResistanceInclinedShearReinforcement`

`equation`

$$
\tau_{Rd} = \min\left(\nu \cdot f_{cd} \cdot \frac{\cot(\theta) - \cot(\beta_{incl})}{1 + \left(\cot(\theta)\right)^2} + \rho_w \cdot f_{ywd} \cdot \left(\cot(\beta_{incl}) + \cot(\alpha_w)\right) \cdot \sin(\alpha_w), \nu \cdot f_{cd} \cdot \frac{\cot(\theta) + \cot(\alpha_w)}{1 + \left(\cot(\theta)\right)^2}\right)
$$

`complete`

$$
\tau_{Rd} = \min\left(\nu \cdot f_{cd} \cdot \frac{\cot(\theta) - \cot(\beta_{incl})}{1 + \left(\cot(\theta)\right)^2} + \rho_w \cdot f_{ywd} \cdot \left(\cot(\beta_{incl}) + \cot(\alpha_w)\right) \cdot \sin(\alpha_w), \nu \cdot f_{cd} \cdot \frac{\cot(\theta) + \cot(\alpha_w)}{1 + \left(\cot(\theta)\right)^2}\right) = \min\left(0.500 \cdot 20.000 \cdot \frac{2.000 - 1.000}{1 + \left(2.000\right)^2} + 0.002 \cdot 435.000 \cdot \left(1.000 + \cot(60.000)\right) \cdot \sin(60.000), 0.500 \cdot 20.000 \cdot \frac{2.000 + \cot(60.000)}{1 + \left(2.000\right)^2}\right) = 3.188 \ MPa
$$

`complete_with_units`

$$
\tau_{Rd} = \min\left(\nu \cdot f_{cd} \cdot \frac{\cot(\theta) - \cot(\beta_{incl})}{1 + \left(\cot(\theta)\right)^2} + \rho_w \cdot f_{ywd} \cdot \left(\cot(\beta_{incl}) + \cot(\alpha_w)\right) \cdot \sin(\alpha_w), \nu \cdot f_{cd} \cdot \frac{\cot(\theta) + \cot(\alpha_w)}{1 + \left(\cot(\theta)\right)^2}\right) = \min\left(0.500 \cdot 20.000 \ MPa \cdot \frac{2.000 - 1.000}{1 + \left(2.000\right)^2} + 0.002 \cdot 435.000 \ MPa \cdot \left(1.000 + \cot(60.000 \ degrees)\right) \cdot \sin(60.000 \ degrees), 0.500 \cdot 20.000 \ MPa \cdot \frac{2.000 + \cot(60.000 \ degrees)}{1 + \left(2.000\right)^2}\right) = 3.188 \ MPa
$$

**Formula (8.63)**, `Form8Dot63StressInInclinedShearReinforcement`

`equation`

$$
\sigma_{swd} = \min\left(E_s \cdot \left[\left(\varepsilon_x + 0.001\right) \cdot \frac{\left(\cot(\theta) + \cot(\alpha_w)\right)^2}{1 + \left(\cot(\alpha_w)\right)^2} - 0.001\right], f_{ywd}\right)
$$

`complete`

$$
\sigma_{swd} = \min\left(E_s \cdot \left[\left(\varepsilon_x + 0.001\right) \cdot \frac{\left(\cot(\theta) + \cot(\alpha_w)\right)^2}{1 + \left(\cot(\alpha_w)\right)^2} - 0.001\right], f_{ywd}\right) = \min\left(200000.000 \cdot \left[\left(0.001 + 0.001\right) \cdot \frac{\left(0.500 + \cot(60.000)\right)^2}{1 + \left(\cot(60.000)\right)^2} - 0.001\right], 435.000\right) = 148.205 \ MPa
$$

`complete_with_units`

$$
\sigma_{swd} = \min\left(E_s \cdot \left[\left(\varepsilon_x + 0.001\right) \cdot \frac{\left(\cot(\theta) + \cot(\alpha_w)\right)^2}{1 + \left(\cot(\alpha_w)\right)^2} - 0.001\right], f_{ywd}\right) = \min\left(200000.000 \ MPa \cdot \left[\left(0.001 + 0.001\right) \cdot \frac{\left(0.500 + \cot(60.000 \ degrees)\right)^2}{1 + \left(\cot(60.000 \ degrees)\right)^2} - 0.001\right], 435.000 \ MPa\right) = 148.205 \ MPa
$$

## Summary by CodeRabbit

- **New Features**
  - Added Eurocode 2 formulas 8.61–8.63 for inclined shear reinforcement.
  - Supports calculating additional tensile force, enhanced shear stress resistance, and reinforcement stress.
  - Added formatted LaTeX output with symbolic, numeric, and unit-inclusive representations.

- **Documentation**
  - Updated the formula status guide to mark formulas 8.61–8.63 as complete.

- **Tests**
  - Added coverage for calculations, validation, boundary conditions, and LaTeX representations.


blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_61.py
```diff
@@ -0,0 +1,86 @@
...
+        The standard takes the magnitude of the shear force, so the sign of [$V_{Ed}$] does not reach the
+        result. The result itself does carry a sign: it turns negative where the shear reinforcement is
+        inclined more steeply than the compression field, that is where [$\cot\alpha_w > \cot\theta$]. The
+        standard prints no bound on it, so that value is returned unchanged.
+
+        Parameters
+        ----------
+        v_ed : N
+            [$V_{Ed}$] Design shear force at the control section. Only its magnitude is used [$N$].
+        cot_theta : DIMENSIONLESS
```
casper-van-der-krift: Not sure if I agree on the decision to treat alpha_w as the 'primary' variable given in degrees, while providing secondary variables regarding theta as cot_theta. I think it's cleaner and more consistent to just pass all angles directly in degrees. If you agree, perhaps include this in the agent's instructions to prevent in future PR's.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_61.py
```diff
@@ -0,0 +1,86 @@
...
+        self.cot_theta = cot_theta
+        self.alpha_w = alpha_w
+
+    @staticmethod
+    def _evaluate(v_ed: N, cot_theta: DIMENSIONLESS, alpha_w: DEG) -> N:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(cot_theta=cot_theta)
+        raise_if_less_or_equal_to_zero(alpha_w=alpha_w)
+
+        return abs(v_ed) * (cot_theta - 1 / np.tan(np.deg2rad(alpha_w)))
```
casper-van-der-krift: While this is mathematically correct, I prefer staying as close the original notation as possible:

```
from utils.math_helpers import cot

return abs(v_ed) * (cot(theta) - cot(alpha_w))
```

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_62.py
```diff
@@ -0,0 +1,137 @@
...
+        self.cot_beta_incl = cot_beta_incl
+        self.rho_w = rho_w
+        self.f_ywd = f_ywd
+        self.alpha_w = alpha_w
+
+    @staticmethod
+    def _evaluate(
+        nu: DIMENSIONLESS,
+        f_cd: MPA,
+        cot_theta: DIMENSIONLESS,
```
casper-van-der-krift: SEC about passing all angles directly as variables and not as cotangens of those angles.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_62.py
```diff
@@ -0,0 +1,137 @@
...
+        rho_w: DIMENSIONLESS,
+        f_ywd: MPA,
+        alpha_w: DEG,
+    ) -> MPA:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(nu=nu, f_cd=f_cd, cot_theta=cot_theta, cot_beta_incl=cot_beta_incl, rho_w=rho_w, f_ywd=f_ywd)
+        raise_if_less_or_equal_to_zero(alpha_w=alpha_w)
+
+        alpha_w_rad = np.deg2rad(alpha_w)
+        cot_alpha_w = 1 / np.tan(alpha_w_rad)
```
casper-van-der-krift: We have our own cotangens function in module `utils.math_helpers`

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_63.py
```diff
@@ -0,0 +1,105 @@
...
+        combination that gives a negative stress returns that value unchanged.
+
+        Parameters
+        ----------
+        e_s : MPA
+            [$E_s$] Modulus of elasticity of the reinforcing steel [$MPa$].
+        epsilon_x : DIMENSIONLESS
+            [$\varepsilon_x$] Longitudinal strain, which may be calculated according to 8.2.3(7) for a
+            cross-section located midway between the support and the load [$-$].
+        cot_theta : DIMENSIONLESS
```
casper-van-der-krift: SEC about all angles being passed directly instead of as cotangens.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_63.py
```diff
@@ -0,0 +1,105 @@
...
+        self.alpha_w = alpha_w
+        self.f_ywd = f_ywd
+
+    @staticmethod
+    def _evaluate(e_s: MPA, epsilon_x: DIMENSIONLESS, cot_theta: DIMENSIONLESS, alpha_w: DEG, f_ywd: MPA) -> MPA:
+        """Evaluates the formula, for more information see the __init__ method."""
+        raise_if_negative(e_s=e_s, epsilon_x=epsilon_x, cot_theta=cot_theta, f_ywd=f_ywd)
+        raise_if_less_or_equal_to_zero(alpha_w=alpha_w)
+
+        cot_alpha_w = 1 / np.tan(np.deg2rad(alpha_w))
```
casper-van-der-krift: SEC about use of `utils.math_helpers.cot`

casper-van-der-krift (changes requested):

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_61.py
egarciamendez: Agreed, changed to `theta: DEG` throughout, and `beta_incl: DEG` in 8.62. Also added to the skill and the verifier agent, so a `cot_theta` parameter now gets flagged before a PR is opened rather than in review.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_61.py
egarciamendez: Applied exactly as written: `return abs(v_ed) * (cot(theta) - cot(alpha_w))`.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_62.py
egarciamendez: Done, `theta: DEG` and `beta_incl: DEG`.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_62.py
egarciamendez: Now using `math_helpers.cot` for all three angles. It rejects angles above 90 degrees, which the standard says should be avoided.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_63.py
egarciamendez: Done, `theta: DEG`.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_63.py
egarciamendez: Done, `cot(theta)` and `cot(alpha_w)` from `math_helpers`.

egarciamendez:
One change in this push was not requested in review. Formula (8.62) had `raise_if_negative(f_ywd=...)` while its own docstring states that 8.2.3(13) requires `f_ywd` to be replaced by the stress of (8.63), which has no printed lower bound and is negative over much of its range. That guard blocked the substitution the standard prescribes. It is the same defect CodeRabbit found on #1099 for (8.57) and (8.56), so I fixed it here rather than leave a known copy behind. Regression test added that feeds the real negative 8.63 result into 8.62.

Note that (8.55) on main still carries the same guard while documenting the same substitution for (8.56). That one is untouched and needs a separate PR.

PR #1103, opened 2026-07-27
feat: add formulas 8.65 and 8.66 from FprEN 1992-1-1:2023 clause 8.2.5

egarciamendez:
Implements formulas (8.65) and (8.66) of FprEN 1992-1-1:2023 (E), clause 8.2.5, paragraphs (1) and (2), pages 128 and 129. This opens clause 8.2.5, shear between web and flanges.

Contributes to #1083.

## A passing check here does not mean the connection is safe

(8.66) is a `ComparisonFormula`, but its meaning is unlike the checks in the earlier pull requests of this series. The standard introduces it with: "In case the following condition is satisfied, further verification of the shear between web and flanges may be omitted and no extra reinforcement above that for transverse bending is required".

So a satisfied check means the designer may skip the work of 8.2.5(3) and (4), not that the junction has been verified. And a failing check does not mean the junction is inadequate: 8.2.5(3) then says the shear strength of the flange "may be calculated by considering the flange as a system of compression fields combined with ties", which routes the design through (8.67) to (8.71) and may well pass.

Both readings are in the class docstring, because `OK` on a check that is really a permission to stop checking is exactly the kind of thing that gets misread downstream. The blind verifier was asked what a satisfied check means before seeing any of this and reached the same two-sided reading from the printed sentences.

## Decisions

- **(8.65) rejects a negative change of axial force, and that is a decision taken here rather than a reading of the standard.** The standard defines `dF_d` only as "the change of the axial force in the flange over the length dx". It prints no sign requirement and no absolute value bars. Letting the sign through would be worse than rejecting: a negative shear stress satisfies (8.66) trivially, for a load case that is just as onerous as its mirror image. Taking the magnitude would also close that path but would silently accept either sign convention. The code comment says all of this, and the verifier independently identified the same failure mode and the same two candidate treatments.
- The limits the standard sets on `dx`, half the distance between the section of zero moment and the section of maximum moment, and no more than the distance between point loads, are conditions on how the caller picks the length. They are not bounds on the result and are not enforced.
- `s_f` is documented as the standard leaves it. The verifier pointed out that neither page defines it in words: it appears only as a dimension in Figure 8.13, between the transverse bars along `dx`. The docstring now says so rather than asserting a definition of its own.

## A boundary test that passed by rounding luck

The first version of the inclusive-bound test used `tau_ed = 4.35` against `400 / (200 * 200) * 435`. That expression evaluates to `4.3500000000000005` while the literal `4.35` is `4.34999999999999964`, so the case passed on rounding rather than on the printed `<=`. It proved nothing about inclusivity.

The verifier flagged it: "Set B is a floating-point knife-edge ... flagging it rather than trusting it." The test now uses `f_yd = 400`, for which `400 / (200 * 200) * 400` is exactly `4.0`, asserts that the right-hand side really is `4.0`, and checks both `4.0` and a value just above it. The realistic `435` stays in the other cases.

## Verification

Both formulas were blindly verified against the rendered pages 128 and 129 of the standard, one agent per formula, neither of which saw the implementation or the tests.

**(8.65): verdict OK.** Same result for all three input sets. It confirmed the unit by working through the dimensions, and that the fraction bar spans both factors of the denominator. On the sign question it was explicit that rejecting a negative value is an addition to the standard and must be documented as the implementer's choice, which is what the code comment now does.

**(8.66): verdict OK.** Same right-hand side and the same verdict for all three input sets, with its arithmetic shown. It quoted the sentence that fixes what a satisfied check means, confirmed the bound is inclusive, and confirmed that `f_yd` multiplies the whole fraction and is not under the bar. Its two remarks, on `s_f` being undefined in words and on the knife-edge boundary case, both changed this pull request.

## Formulas as implemented

**Formula (8.65)**, `Form8Dot65LongitudinalShearStressFlangeWebJunction`

`equation`

$$
\tau_{Ed} = \frac{\Delta F_d}{h_f \cdot \Delta x}
$$

`complete`

$$
\tau_{Ed} = \frac{\Delta F_d}{h_f \cdot \Delta x} = \frac{200000.000}{200.000 \cdot 2000.000} = 0.500 \ MPa
$$

`complete_with_units`

$$
\tau_{Ed} = \frac{\Delta F_d}{h_f \cdot \Delta x} = \frac{200000.000 \ N}{200.000 \ mm \cdot 2000.000 \ mm} = 0.500 \ MPa
$$

**Formula (8.66)**, `Form8Dot66CheckOmissionOfShearVerification`

`equation`

$$
CHECK \to \tau_{Ed} \leq \frac{A_{st,min}}{s_f \cdot h_f} \cdot f_{yd}
$$

`complete`

$$
CHECK \to \tau_{Ed} \leq \frac{A_{st,min}}{s_f \cdot h_f} \cdot f_{yd} \to 0.500 \leq \frac{400.000}{200.000 \cdot 200.000} \cdot 435.000 \to OK
$$

`complete_with_units`

$$
CHECK \to \tau_{Ed} \leq \frac{A_{st,min}}{s_f \cdot h_f} \cdot f_{yd} \to 0.500 \ MPa \leq \frac{400.000 \ mm^2}{200.000 \ mm \cdot 200.000 \ mm} \cdot 435.000 \ MPa \to OK
$$

## Summary by CodeRabbit

* **New Features**
  * Added Eurocode 2:2023 Formula 8.65 for calculating longitudinal shear stress at a flange–web junction.
  * Added Formula 8.66 to verify when shear verification may be omitted based on reinforcement and stress limits.
  * Both formulas support validation, numerical results, and LaTeX representations with units.

* **Documentation**
  * Updated the Eurocode formula status table to mark Formulas 8.65 and 8.66 as complete.

* **Tests**
  * Added coverage for calculations, validation, threshold behavior, comparison results, and LaTeX output.

### DIFF
--- a/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_66.py
+++ b/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_66.py
@@ -1,0 +1,106 @@
+"""Formula 8.66 from FprEN 1992-1-1:2023: Chapter 8: Ultimate limit states (ULS)."""
+
+import operator
+from collections.abc import Callable
+
+from blueprints.codes.eurocode.fpr_en_1992_1_1_2023 import FPR_EN_1992_1_1_2023
+from blueprints.codes.formula import ComparisonFormula
+from blueprints.codes.latex_formula import LatexFormula, latex_replace_symbols
+from blueprints.type_alias import MM, MM2, MPA
+from blueprints.validations import raise_if_less_or_equal_to_zero, raise_if_negative
+
+
+class Form8Dot66CheckOmissionOfShearVerification(ComparisonFormula):
+    r"""Class representing formula 8.66 for the check that determines whether the verification of the shear
+    between web and flanges may be omitted.
+
+    A satisfied check does not mean the connection has been verified. It means the standard permits the
+    verification to be skipped and requires no extra reinforcement above that for transverse bending. A check
+    that is not satisfied does not mean the connection fails either; it means 8.2.5(3) and (4) have to be
+    worked through, starting from Formula (8.69).
+    """
+
+    label = "8.66"
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(self, tau_ed: MPA, a_st_min: MM2, s_f: MM, h_f: MM, f_yd: MPA) -> None:
+        r"""Check whether further verification of the shear between web and flanges may be omitted.
+
+        FprEN 1992-1-1:2023 (E) art 8.2.5 (2) - Formula (8.66)
+
+        Parameters
+        ----------
+        tau_ed : MPA
+            [$\tau_{Ed}$] Longitudinal shear stress at the junction between one side of a flange and the web
+            according to Formula (8.65), see Form8Dot65LongitudinalShearStressFlangeWebJunction [$MPa$].
+        a_st_min : MM2
+            [$A_{st,min}$] Minimum transverse reinforcement according to Table 12.1 (NDP) [$mm^2$].
+        s_f : MM
+            [$s_f$] Spacing of the transverse reinforcement in the flange. The standard does not define it in
+            words; it appears only as a dimension in Figure 8.13, between the transverse bars along
+            [$\Delta x$] [$mm$].
+        h_f : MM
+            [$h_f$] Thickness of the flange at the junctions [$mm$].
+        f_yd : MPA
+            [$f_{yd}$] Design value of the yield strength of the transverse reinforcement [$MPa$].
+        """
+        super().__init__()
+        self.tau_ed = tau_ed
+        self.a_st_min = a_st_min
+        self.s_f = s_f
+        self.h_f = h_f
+        self.f_yd = f_yd
+
+    @classmethod
+    def _comparison_operator(cls) -> Callable[[float, float], bool]:
+        return operator.le
+
+    @staticmethod
+    def _evaluate_lhs(tau_ed: MPA, *_args, **_kwargs) -> float:
+        """Evaluates the longitudinal shear stress, for more information see the __init__ method."""
+        raise_if_negative(tau_ed=tau_ed)
+
+        return float(tau_ed)
+
+    @staticmethod
+    def _evaluate_rhs(a_st_min: MM2, s_f: MM, h_f: MM, f_yd: MPA, *_args, **_kwargs) -> float:
+        """Evaluates the stress that the minimum transverse reinforcement already carries."""
+        raise_if_negative(a_st_min=a_st_min, f_yd=f_yd)
+        raise_if_less_or_equal_to_zero(s_f=s_f, h_f=h_f)
+
+        return float(a_st_min / (s_f * h_f) * f_yd)
+
+    def latex(self, n: int = 3) -> LatexFormula:
+        """Returns LatexFormula object for formula 8.66."""
+        _equation: str = r"\tau_{Ed} \leq \frac{A_{st,min}}{s_f \cdot h_f} \cdot f_{yd}"
+        _numeric_equation: str = latex_replace_symbols(
+            template=_equation,
+            replacements={
+                r"\tau_{Ed}": f"{self.tau_ed:.{n}f}",
+                r"A_{st,min}": f"{self.a_st_min:.{n}f}",
+                r"s_f": f"{self.s_f:.{n}f}",
+                r"h_f": f"{self.h_f:.{n}f}",
+                r"f_{yd}": f"{self.f_yd:.{n}f}",
+            },
+            unique_symbol_check=False,
+        )
+        _numeric_equation_with_units: str = latex_replace_symbols(
+            template=_equation,
+            replacements={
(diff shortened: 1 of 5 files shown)


blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_66.py
```diff
@@ -0,0 +1,106 @@
...
+    def _evaluate_lhs(tau_ed: MPA, *_args, **_kwargs) -> float:
+        """Evaluates the longitudinal shear stress, for more information see the __init__ method."""
+        raise_if_negative(tau_ed=tau_ed)
+
+        return float(tau_ed)
+
+    @staticmethod
+    def _evaluate_rhs(a_st_min: MM2, s_f: MM, h_f: MM, f_yd: MPA, *_args, **_kwargs) -> float:
+        """Evaluates the stress that the minimum transverse reinforcement already carries."""
+        raise_if_negative(a_st_min=a_st_min, f_yd=f_yd)
```
rhhk-dot, line 68: Would exclude zero for a_st_min and f_yd as well. See comment CodeRabbit.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_66.py
```diff
@@ -0,0 +1,106 @@
...
+        raise_if_less_or_equal_to_zero(s_f=s_f, h_f=h_f)
+
+        return float(a_st_min / (s_f * h_f) * f_yd)
+
+    def latex(self, n: int = 3) -> LatexFormula:
```
rhhk-dot, line 73: Add intermediate result for rhs

tests/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/test_formula_8_66.py
```diff
@@ -0,0 +1,118 @@
...
+        [
+            # the minimum transverse reinforcement carries 400 / (200 * 200) * 435 = 4.35 MPa
+            (0.5, True, 0.11494252873563217),
+            (5.0, False, 1.1494252873563218),
+        ],
+    )
+    def test_evaluation(self, tau_ed: float, expected: bool, expected_unity_check: float) -> None:
+        """Tests the verdict and the unity check that it is based on."""
+        # Create object to test
+        formula = Form8Dot66CheckOmissionOfShearVerification(tau_ed=tau_ed, a_st_min=400.0, s_f=200.0, h_f=200.0, f_yd=435.0)
```
rhhk-dot, line 25: Why not parametrize a_st_min to f_yd as well?

rhhk-dot (changes requested):
Reviewed

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_66.py
egarciamendez: a_st_min and f_yd now go through raise_if_less_or_equal_to_zero instead of raise_if_negative.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_66.py
egarciamendez: latex() now includes an intermediate_result showing the evaluated lhs/rhs, e.g. "0.500 \leq 4.350", before the OK/Not OK.

tests/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/test_formula_8_66.py
egarciamendez: Added cases varying a_st_min and f_yd, including one that flips the verdict.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_66.py
```diff
@@ -0,0 +1,106 @@
...
+
+        return float(tau_ed)
+
+    @staticmethod
+    def _evaluate_rhs(a_st_min: MM2, s_f: MM, h_f: MM, f_yd: MPA, *_args, **_kwargs) -> float:
+        """Evaluates the stress that the minimum transverse reinforcement already carries."""
+        raise_if_negative(a_st_min=a_st_min, f_yd=f_yd)
+        raise_if_less_or_equal_to_zero(s_f=s_f, h_f=h_f)
+
+        return float(a_st_min / (s_f * h_f) * f_yd)
```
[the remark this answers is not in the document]
egarciamendez: Fixed the zero-capacity part, a_st_min and f_yd now rejected at zero. Kept the positional-argument shape as is, every ComparisonFormula in this repo has it and every call site uses keyword arguments only.

PR #1105, opened 2026-07-27
fix: derive the boolean of a comparison formula from its operator instead of its unity check

egarciamendez:
`ComparisonFormula.__bool__` derived its answer from `unity_check`, the ratio of the two sides. That ratio only orders them correctly while both have the same sign, so a comparison whose sides straddle zero was reported wrongly, and one with a right-hand side of zero raised instead of answering.

`ComparisonFormula._evaluate` never had the problem: it applies the operator. So `float(formula)` was right while `bool(formula)` was wrong, and the two disagreed silently.

## What was wrong

With `lhs = 2`, `rhs = -2` and the `le` operator, `2 <= -2` is false, but `2 / -2 = -1` and `-1 <= 1` is true, so the check reported satisfied. With `lhs = -2`, `rhs = -1` it went the other way: `-2 <= -1` is true, but the ratio is `2` and the check reported violated. With `rhs = 0` there is no ratio at all and it raised `ZeroDivisionError`.

This is reachable in ordinary use. A lower bound has to be written as `constant <= value`, which puts the value under check on the right-hand side. Three range checks from the FprEN 1992-1-1:2023 series are built that way and all three were affected:

```
8.41  cot_theta = -1.0                 -> reported OK,  while 1 <= -1.0 <= 2.5 is false
8.58  cot_theta = -1.0, alpha_w = 60   -> reported OK,  while 0.577 <= -1.0 <= 2.5 is false
8.67  cot_theta_f = -1.0, compression  -> reported OK,  while 1 <= -1.0 <= 3.0 is false
all three, cot = 0.0                   -> ZeroDivisionError
```

`AggregatedComparisonFormula` inherited the same `__bool__` while disabling the comparison operator and both sides, so it could not be corrected by the same change alone. Its own `_evaluate` aggregates over `bool(member)` and was therefore correct all along, which is the second disagreement this fixes.

## The fix

Two methods, no change to any signature or to `unity_check`:

- `ComparisonFormula.__bool__` applies the comparison operator to the two sides.
- `AggregatedComparisonFormula.__bool__` applies the aggregation function to the results of its members, which is exactly what its `_evaluate` already did.

After this, `bool(formula)` and `float(formula)` agree by construction in both classes.

## Evidence

The whole suite passes: **6760 tests, no regressions**. That covers the 24 comparison formulas already in the repository, none of which changed behaviour, because their two sides are same-signed in every case they are tested with.

Six new cases in `tests/codes/test_formula.py` pin the behaviour down, covering both signs on either side, the zero right-hand side, and the two aggregations. Each of them also asserts that `bool()` and `float()` agree, which is the invariant that was broken.

## What this does not change

`unity_check` is left as it is. The ratio is still what a reader wants for the ordinary same-signed case, and it is what the aggregations report. It remains ill-defined where the two sides straddle zero, but it no longer decides whether a check passes, which is what made that mattering.

Worth a separate look, not touched here: for an `any` aggregation `unity_check` takes the smallest member ratio, which is not the same question as whether any of the conditions holds. With this fix the boolean no longer depends on it, so the two can now disagree in that case.

### DIFF
--- a/blueprints/codes/formula.py
+++ b/blueprints/codes/formula.py
@@ -211,21 +211,28 @@ class ComparisonFormula(Formula, ABC):
     def __bool__(self) -> bool:
         """Return whether the comparison condition is satisfied.
 
-        Returns True if the unity check is less than or equal to 1.0, indicating the condition is satisfied.
-        This allows ComparisonFormula instances to be used directly in boolean contexts.
+        The comparison operator is applied to the two sides directly, so that this agrees with
+        :meth:`_evaluate` and therefore with ``float(self)`` for every input.
+
+        It deliberately does not go through :attr:`unity_check`. That ratio only orders the two sides
+        correctly while both have the same sign. Where they do not, ``lhs / rhs`` turns negative and a
+        negative value is below 1, so a condition that is in fact violated would be reported as satisfied;
+        a right-hand side of zero would raise :class:`ZeroDivisionError` instead of answering. Both are
+        reachable whenever a lower bound is written as ``constant <= value``, which puts the value under
+        check on the right-hand side.
 
         Examples
         --------
         formula = SomeComparisonFormula(...)
-        if formula:  # Equivalent to: if formula.unity_check <= 1.0
+        if formula:  # Equivalent to: if formula.lhs <= formula.rhs, for the le operator
             print("Condition satisfied")
 
         Returns
         -------
         bool
-            True if unity_check <= 1.0 (condition is satisfied), False otherwise.
+            True if the comparison holds, False otherwise.
         """
-        return self.unity_check <= 1.0
+        return bool(self._comparison_operator()(self.lhs, self.rhs))
 
     @classmethod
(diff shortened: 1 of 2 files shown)


egarciamendez:
Closing in favour of #1108, which fixes the same two methods the same way and carries more with it.

Same diagnosis, same change: `ComparisonFormula.__bool__` applies the comparison operator, and `AggregatedComparisonFormula` gets its own `__bool__` that aggregates over its members instead of inheriting the unity check path. On top of that #1108 fixes the unity check in 5.7, 5.8, 5.18 and 5.19 for #1107, and gives `AggregatedComparisonFormula` a default `latex()`. Its tests also cover the sign crossing cases for both `le` and `ge`, where this branch only covered `le`.

Two pull requests changing the same lines of `formula.py` would only collide, and #1108 is the broader one.

The one thing from here that is not in #1108 is the `bool()` cast around the operator result, which keeps `__bool__` working when the two sides are numpy floats. Left as an inline comment there rather than kept as a competing branch.

PR #1106, opened 2026-07-27
feat: add formulas 8.67 and 8.68 from FprEN 1992-1-1:2023 clause 8.2.5

egarciamendez:
Implements formulas (8.67) and (8.68) of FprEN 1992-1-1:2023 (E), clause 8.2.5(3), page 129.

Contributes to #1083.

## A defect this pull request found, which also sits in #1096 and #1100

**A two-sided range check built on `AggregatedComparisonFormula` reports OK for a negative value under check, and raises `ZeroDivisionError` for zero.**

`ComparisonFormula.__bool__` answers through the unity check, which for `operator.le` is `lhs / rhs`. A lower bound has to be written as `constant <= value`, so the value under check sits in `rhs`. A negative value flips the sign of that ratio, and `-1.0 <= 1.0` is true, so the bound reports OK. Note that `ComparisonFormula._evaluate` compares `lhs` and `rhs` with the operator directly and gets it right, so `float(formula)` is correct while `bool(formula)` is not; `AggregatedComparisonFormula._evaluate` aggregates over `bool(...)`, so the aggregate inherits the wrong answer.

Measured on the three range checks in this series:

```
8.41  cot_theta = -1.0                 -> reports OK,  while 1 <= -1.0 <= 2.5 is False
8.58  cot_theta = -1.0, alpha_w = 60   -> reports OK,  while 0.577 <= -1.0 <= 2.5 is False
8.67  cot_theta_f = -1.0, compression  -> reports OK,  while 1 <= -1.0 <= 3.0 is False
all three, cot = 0.0                   -> ZeroDivisionError
```

This did not exist before: `DoubleComparisonFormula` applied its operators directly, and the defect arrived with the move to `AggregatedComparisonFormula`.

**The root cause is fixed in #1105**, which makes both `ComparisonFormula.__bool__` and `AggregatedComparisonFormula.__bool__` apply their operator instead of the ratio. That closes it for every comparison formula, present and future, and for values of any sign. Once it lands, #1096 and #1100 are correct without any change of their own; until then they carry the defect.

**This pull request also guards the cotangent**, and keeps that guard whatever happens to #1105. A cotangent of zero or less is not the inclination of a compression field, so it is refused as input rather than answered as a check. Both halves refuse it on their own, not only through the aggregate, and there are tests for both. After #1105 the guard is no longer what makes the class correct, it is only what keeps meaningless input out.

## Decisions

- The two formulas are one class. They sit under a single introductory sentence in 8.2.5(3), share the lower bound of 1, share the inclusivity of both bounds, and differ only in the upper bound: 3,0 in compression flanges, 1,25 in tension flanges. The flange state is therefore an input, the same way (8.22) to (8.24) are one class selected by a ratio. The blind verifier was asked whether that modelling is faithful or whether the standard means two different things, and answered that it is one selectable range whose ceiling depends on the state of the flange.
- The flange state is a `Literal["compression", "tension"]` mapped to the limit, following `Form5Dot1CriteriumDisregardSecondOrderEffects`. Anything else raises, because the page offers no third case and no default.
- 8.2.5(5) allows lower angles in the tensile flange than (8.68) gives, provided the factor `nu` is calculated from the state of strains rather than taken as the flat 0,5 of (8.71). The verifier confirmed that this is a separate design route tied to the crushing check and not a relaxation of this one, so the 1,25 cap stays hard here.
- The equation keeps the word `limit` as a placeholder that the flange state fills in, which is what `Form5Dot1` does for the same situation.

## Verification

Blindly verified against the rendered pages 128 and 129 of the standard by a second agent that never saw the implementation or the tests.

Verdict **OK**. It independently derived clause 8.2.5(3), both bounds and their inclusivity for each of the two formulas, and the same boolean for all seven test input sets, including the discriminating pair where the same cotangent passes in a compression flange and fails in a tension flange. It confirmed the two differ in nothing but the upper bound, confirmed the one-class modelling, and read the 8.2.5(5) route correctly as separate. Its remark on the unity check of the lower bound is what led to the defect above.

## Formulas as implemented

**Formulas (8.67) and (8.68)**, `Form8Dot67To68CheckCotangentFlangeCompressionField`

`equation`

$$
CHECK \to 1 \leq \cot(\theta_f) \leq limit
$$

`complete`, with `cot_theta_f = 2,0` in a compression flange, where the limit is 3,0

$$
CHECK \to 1 \leq \cot(\theta_f) \leq limit \to 1 \leq 2.000 \leq 3.000 \to OK
$$

and the same cotangent in a tension flange, where the limit is 1,25

$$
CHECK \to 1 \leq \cot(\theta_f) \leq limit \to 1 \leq 2.000 \leq 1.250 \to \text{Not OK}
$$

`complete_with_units` is deliberately the same string as `complete`, since every term here is dimensionless.

## Summary by CodeRabbit

* **New Features**
  * Added support for Eurocode 2 formulas 8.67 and 8.68, checking the cotangent of the flange compression-field inclination.
  * Supports separate limits for compression and tension flanges, with validation for invalid angles and flange types.
  * Added readable formula representations, including versions with substituted values and units.

* **Documentation**
  * Updated the Eurocode 2 formula tracking table to mark formulas 8.67 and 8.68 as implemented.

### DIFF
--- a/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
+++ b/blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
@@ -1,0 +1,222 @@
+"""Formula 8.67 and 8.68 from FprEN 1992-1-1:2023: Chapter 8: Ultimate limit states (ULS)."""
+
+import operator
+from collections.abc import Callable, Sequence
+from typing import Literal, Self
+
+from blueprints.codes.eurocode.fpr_en_1992_1_1_2023 import FPR_EN_1992_1_1_2023
+from blueprints.codes.formula import AggregatedComparisonFormula, ComparisonFormula
+from blueprints.codes.latex_formula import LatexFormula, latex_replace_symbols
+from blueprints.type_alias import DIMENSIONLESS
+from blueprints.validations import raise_if_less_or_equal_to_zero
+
+FLANGE_TYPE = Literal["compression", "tension"]
+
+
+def _upper_bound(flange_type: FLANGE_TYPE) -> float:
+    """Returns the upper bound on the cotangent, which is 3,0 in compression flanges per Formula (8.67) and
+    1,25 in tension flanges per Formula (8.68).
+    """
+    limits = {"compression": 3.0, "tension": 1.25}
+    limit = limits.get(flange_type.lower())
+    if limit is None:
+        raise ValueError(f"Invalid flange type: {flange_type}. Must be 'compression' or 'tension'.")
+    return limit
+
+
+class SubForm8Dot67To68LowerBound(ComparisonFormula):
+    r"""Class representing the lower bound of formulas 8.67 and 8.68, [$1 \leq \cot\theta_f$].
+
+    The standard prints the same lower bound for both flange types.
+    """
+
+    label = "8.67/8.68"
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(self, cot_theta_f: DIMENSIONLESS) -> None:
+        r"""Check the selected inclination of the compression field in the flange against its lower bound.
+
+        FprEN 1992-1-1:2023 (E) art 8.2.5 (3) - Formula (8.67) and (8.68)
+
+        Parameters
+        ----------
+        cot_theta_f : DIMENSIONLESS
+            [$\cot\theta_f$] Cotangent of the selected inclination of the compression field in the flange
+            with respect to the longitudinal axis [$-$].
+        """
+        super().__init__()
+        self.cot_theta_f = cot_theta_f
+
+    @classmethod
+    def _comparison_operator(cls) -> Callable[[float, float], bool]:
+        return operator.le
+
+    @staticmethod
+    def _evaluate_lhs(*_args, **_kwargs) -> float:
+        """Evaluates the lower bound, which the standard prints as the constant 1."""
+        return 1.0
+
+    @staticmethod
+    def _evaluate_rhs(cot_theta_f: DIMENSIONLESS, *_args, **_kwargs) -> float:
+        """Evaluates the value under check, for more information see the __init__ method."""
+        # A cotangent of zero or less is not the inclination of a compression field, and it must be refused
+        # rather than reported as a failed check. ComparisonFormula.__bool__ answers through the unity check,
+        # which for a lower bound written "constant <= value" is constant/value. A negative value flips the
+        # sign of that ratio, so the bound would silently report OK, and a zero value divides by zero.
+        raise_if_less_or_equal_to_zero(cot_theta_f=cot_theta_f)
+
+        return float(cot_theta_f)
+
+    def latex(self, n: int = 3) -> LatexFormula:
+        """Returns LatexFormula object for the lower bound."""
+        _equation: str = r"1 \leq \cot(\theta_f)"
+        return LatexFormula(
+            return_symbol=r"CHECK",
+            result="OK" if self.__bool__() else r"\text{Not OK}",
+            equation=_equation,
+            numeric_equation=latex_replace_symbols(
+                template=_equation,
+                replacements={r"\cot(\theta_f)": f"{self.cot_theta_f:.{n}f}"},
+                unique_symbol_check=False,
+            ),
+            comparison_operator_label=r"\to",
+            unit="",
+        )
+
+
+class SubForm8Dot67To68UpperBound(ComparisonFormula):
+    r"""Class representing the upper bound of formulas 8.67 and 8.68, [$\cot\theta_f \leq 3,0$] in compression
+    flanges and [$\cot\theta_f \leq 1,25$] in tension flanges.
+    """
+
+    label = "8.67/8.68"
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(self, cot_theta_f: DIMENSIONLESS, flange_type: FLANGE_TYPE) -> None:
+        r"""Check the selected inclination of the compression field in the flange against its upper bound.
+
+        FprEN 1992-1-1:2023 (E) art 8.2.5 (3) - Formula (8.67) and (8.68)
+
+        Parameters
+        ----------
+        cot_theta_f : DIMENSIONLESS
+            [$\cot\theta_f$] Cotangent of the selected inclination of the compression field in the flange [$-$].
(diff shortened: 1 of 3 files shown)


blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
```diff
@@ -0,0 +1,222 @@
...
+from collections.abc import Callable, Sequence
+from typing import Literal, Self
+
+from blueprints.codes.eurocode.fpr_en_1992_1_1_2023 import FPR_EN_1992_1_1_2023
+from blueprints.codes.formula import AggregatedComparisonFormula, ComparisonFormula
+from blueprints.codes.latex_formula import LatexFormula, latex_replace_symbols
+from blueprints.type_alias import DIMENSIONLESS
+from blueprints.validations import raise_if_less_or_equal_to_zero
+
+FLANGE_TYPE = Literal["compression", "tension"]
```
rhhk-dot, line 13: I prefer using Enums in this case, but discussed earlier and OK this way :)

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
```diff
@@ -0,0 +1,222 @@
...
+    1,25 in tension flanges per Formula (8.68).
+    """
+    limits = {"compression": 3.0, "tension": 1.25}
+    limit = limits.get(flange_type.lower())
+    if limit is None:
+        raise ValueError(f"Invalid flange type: {flange_type}. Must be 'compression' or 'tension'.")
+    return limit
+
+
+class SubForm8Dot67To68LowerBound(ComparisonFormula):
```
rhhk-dot, line 27: Formula is not intended to use separately, so could be private.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
```diff
@@ -0,0 +1,222 @@
...
+    r"""Class representing the lower bound of formulas 8.67 and 8.68, [$1 \leq \cot\theta_f$].
+
+    The standard prints the same lower bound for both flange types.
+    """
+
+    label = "8.67/8.68"
```
rhhk-dot, line 33: Would have been fine to split these as separate formulas.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
```diff
@@ -0,0 +1,222 @@
...
+    source_document = FPR_EN_1992_1_1_2023
+
+    def __init__(self, cot_theta_f: DIMENSIONLESS) -> None:
```
rhhk-dot, line 36: For formula 8.41 we use angle theta instead of cot_theta. I prefer consistent usage.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
```diff
@@ -0,0 +1,222 @@
...
+                template=_equation,
+                replacements={r"\cot(\theta_f)": f"{self.cot_theta_f:.{n}f}"},
+                unique_symbol_check=False,
+            ),
+            comparison_operator_label=r"\to",
+            unit="",
+        )
+
+
+class SubForm8Dot67To68UpperBound(ComparisonFormula):
```
rhhk-dot, line 87: Naming `Form8Dot67And68` instead of `Form8Dot67To68` makes more sense imo, since it are just two formulas and not a range of formulas > 2.

rhhk-dot (changes requested):
Reviewed

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
egarciamendez: SubForm are already being used a while ago

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
egarciamendez: Fixed, now takes theta_f (DEG) and computes cot() internally, matching 8.41, 8.58, 8.60, 8.69, 8.70.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
egarciamendez: Renamed to `Form8Dot67And68CheckCotangentFlangeCompressionField` (and the sub-bound classes to match).

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
egarciamendez: Kept public, matches the SubForm8DotXX pattern in 8.58 and 8.85.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
egarciamendez: Kept as one class, same tradeoff as (8.22)-(8.24): one class selected by a ratio or state.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_67_68.py
egarciamendez: Kept as `Literal`, per our earlier discussion.

PR #1108, opened 2026-07-28
1107 bug fix bool operator for comparison formula

### LINKED ISSUE #1107: [🐞 Bug]: Fix bool operator for comparison formula
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/blue-prints/) of Blueprints.

### Issue description

\_\_bool\_\_ should use _comparison_operator

### Reproducible example

```python

```

### Expected behavior

_No response_

### Installed versions

```

```

SZeltaat:
## Description

Hi Enrique,
This is the fix (Sina edition). My PR text is not as fancy as yours, but my code is human written and checked. 😄 

The unity check for formula 5-7 and 5-8 was wrong. When you have lhs > rhs, you should calculate the u.c. = rhs / lhs. 

Fixes #1107 
## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected how comparison-based formulas determine satisfaction (including proper handling for negative values and real boolean results).
  * Updated unity-check calculations for sway imperfections and slenderness-related checks to match the intended expressions.
  * Improved aggregated results output so LaTeX clearly renders **“OK”** vs **“Not OK”** consistently.
* **Tests**
  * Expanded regression coverage for comparison truthiness and aggregated LaTeX rendering.
  * Updated expected values to reflect corrected unity-check and comparison behavior.

### DIFF
--- a/blueprints/codes/formula.py
+++ b/blueprints/codes/formula.py
@@ -225,7 +225,9 @@ class ComparisonFormula(Formula, ABC):
         bool
             True if the comparison condition is satisfied, False otherwise.
         """
-        return self._comparison_operator()(self.lhs, self.rhs)
+        # bool() is required here: if _evaluate_lhs/_evaluate_rhs return numpy floats, the comparison
+        # operator returns a numpy.bool_, and Python requires __bool__ to return a real bool.
+        return bool(self._comparison_operator()(self.lhs, self.rhs))
 
     @classmethod
     def _evaluate(cls, *args, **kwargs) -> bool:
--- a/blueprints/codes/formula.py
+++ b/blueprints/codes/formula.py
@@ -211,7 +211,7 @@ class ComparisonFormula(Formula, ABC):
     def __bool__(self) -> bool:
         """Return whether the comparison condition is satisfied.
 
-        Returns True if the unity check is less than or equal to 1.0, indicating the condition is satisfied.
+        Returns True if the comparison e.g. lhs <= rhs is satisfied.
         This allows ComparisonFormula instances to be used directly in boolean contexts.
 
         Examples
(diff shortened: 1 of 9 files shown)


blueprints/codes/formula.py
```diff
@@ -211,29 +211,29 @@ def unity_check(self) -> float:
...
             print("Condition satisfied")
 
         Returns
         -------
         bool
-            True if unity_check <= 1.0 (condition is satisfied), False otherwise.
+            True if the comparison condition is satisfied, False otherwise.
         """
-        return self.unity_check <= 1.0
+        return self._comparison_operator()(self.lhs, self.rhs)
```
egarciamendez: One thing worth adding: wrap this in `bool()`.

If `_evaluate_lhs` returns a numpy float, `operator.le` gives back a `numpy.bool`, and Python requires `__bool__` to return a real `bool`:

```
operator.le(np.float64(1.0), 2.0)  ->  numpy.bool

without bool() cast : TypeError: __bool__ should return bool, returned numpy.bool
with    bool() cast : True
```

This is not hypothetical: it is what `Form8Dot41InclinationCompressionField` hit before the `float()` wrappers went into `math_helpers`, and what (8.60) hit for me. Those wrappers cover it as long as everyone goes through `cot`, `sec` and `csc`, but a formula that reaches for `np.sqrt` or `np.tan` directly in `_evaluate_lhs` brings it straight back.

`AggregatedComparisonFormula.__bool__` does not need it, since `all()` and `any()` already return a real bool.

egarciamendez (changes requested):

blueprints/codes/formula.py
SZeltaat: Good point! I could not really think of a situation that bool would be necessary.
Made the changes and added tests for this edge case

egarciamendez (approved):

PR #1120, opened 2026-08-07
feat: add formulas 8.69, 8.70 and 8.71 from FprEN 1992-1-1:2023 clause 8.2.5

egarciamendez:
Implements clause 8.2.5(4) of FprEN 1992-1-1:2023 (E), page 129: the transverse reinforcement in the flange and the crushing of the compression field in it.

- Formula (8.69), `Form8Dot69CheckTransverseReinforcementInFlange`. The standard introduces the paragraph as a way to determine `A_sf`, but the line it prints is a relation the longitudinal shear stress has to satisfy, so it is a `ComparisonFormula` and not a formula returning an area.
- Formula (8.70), `Form8Dot70CheckCrushingOfCompressionFieldInFlange`. The printed line both defines the compressive stress and imposes the limit on it. The paragraph introduces it with "the following condition should be satisfied", so it is a verification and not a value with a cap; the stress itself stays available through the `lhs` property.
- Formula (8.71), `Form8Dot71StrengthReductionFactor`. A printed constant, implemented as a class without arguments so that the value carries the same label, source reference and representation as every other numbered formula, and so that calling code can pass it to (8.70) instead of a loose 0,5.

Both checks take `cot_theta_f` rather than an angle in degrees, consistent with Formulas (8.67) and (8.68) in #1106, which bound that same cotangent. In (8.70) the printed `tan` is taken as its reciprocal, so no second angle input is needed.

### Notes

- (8.69) and (8.70) carry no flange-type flag, so neither can enforce the tighter tension-flange bound of Formula (8.68) on `cot_theta_f`. That is deliberate: 8.2.5(3) is a separate numbered check, implemented in #1106, and repeating it inside these two would report the same failure twice.
- (8.69) is phrased in the standard as determining `A_sf` while it is implemented as a check. Stated in the class docstring.
- (8.71) belongs to paragraph (4) and not (5); paragraph (5) starts on the line below it.

Deliberate deviations from the printed page:

- `\cot(\theta_f)` and `\tan(\theta_f)` are written with parentheses where the standard prints `cot θ_f` bare. This follows the existing convention in this repository.
- (8.71) renders as `0.5`, with the decimal point this repository uses, where the standard prints `0,5`.
- The dimensionless unit of (8.71) renders as a lone minus sign after the result, which is the existing convention across the repository.

Contributes to #1083.

## Formulas as implemented

**Formula (8.69)**, `Form8Dot69CheckTransverseReinforcementInFlange`

`equation`

$$
CHECK \to \tau_{Ed} \leq \frac{A_{sf}}{s_f \cdot h_f} \cdot f_{yd} \cdot \cot(\theta_f)
$$

`complete`

$$
CHECK \to \tau_{Ed} \leq \frac{A_{sf}}{s_f \cdot h_f} \cdot f_{yd} \cdot \cot(\theta_f) \to 1.500 \leq \frac{201.000}{150.000 \cdot 200.000} \cdot 435.000 \cdot 1.200 \to OK
$$

`complete_with_units`

$$
CHECK \to \tau_{Ed} \leq \frac{A_{sf}}{s_f \cdot h_f} \cdot f_{yd} \cdot \cot(\theta_f) \to 1.500 \ MPa \leq \frac{201.000 \ mm^2}{150.000 \ mm \cdot 200.000 \ mm} \cdot 435.000 \ MPa \cdot 1.200 \to OK
$$

**Formula (8.70)**, `Form8Dot70CheckCrushingOfCompressionFieldInFlange`

`equation`

$$
CHECK \to \sigma_{cd} = \tau_{Ed} \left(\cot(\theta_f) + \tan(\theta_f)\right) \leq \nu \cdot f_{cd}
$$

`complete`

$$
CHECK \to \sigma_{cd} = \tau_{Ed} \left(\cot(\theta_f) + \tan(\theta_f)\right) \leq \nu \cdot f_{cd} \to 6.100 = 3.000 \left(1.200 + 0.833\right) \leq 0.500 \cdot 20.000 \to OK
$$

`complete_with_units`

$$
CHECK \to \sigma_{cd} = \tau_{Ed} \left(\cot(\theta_f) + \tan(\theta_f)\right) \leq \nu \cdot f_{cd} \to 6.100 \ MPa = 3.000 \ MPa \left(1.200 + 0.833\right) \leq 0.500 \cdot 20.000 \ MPa \to OK
$$

**Formula (8.71)**, `Form8Dot71StrengthReductionFactor`

`equation`

$$
\nu = 0.5
$$

`complete`

$$
\nu = 0.5 = 0.500 \ -
$$

`complete_with_units`

$$
\nu = 0.5 = 0.500 \ -
$$

## Summary by CodeRabbit

* **New Features**
  * Added Eurocode checks for transverse flange reinforcement and prevention of compression-field crushing.
  * Flange compression-field angles are entered in degrees, with automatic trigonometric calculations and validation.
  * Added the flange compression-field strength reduction factor, valued at 0.5.
  * Added numeric, unit-aware, and LaTeX representations for formulas 8.69–8.71.
* **Tests**
  * Added coverage for calculations, boundary conditions, invalid angles, and rendered outputs.


blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_70.py
```diff
@@ -0,0 +1,114 @@
...
+        raise_if_less_or_equal_to_zero(cot_theta_f=cot_theta_f)
+
+        return float(tau_ed * (cot_theta_f + 1 / cot_theta_f))
+
+    @staticmethod
+    def _evaluate_rhs(nu: DIMENSIONLESS, f_cd: MPA, *_args, **_kwargs) -> float:
+        """Evaluates the compressive strength of the compression field, for more information see the __init__ method."""
+        raise_if_negative(nu=nu, f_cd=f_cd)
+
+        return float(nu * f_cd)
```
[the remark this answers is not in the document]
egarciamendez: Zero: fixed in aaaf5212, for `nu` and `f_cd` here and for `f_yd` and `cot_theta_f` in 8.69. `a_sf` still accepts zero, since a flange without transverse reinforcement is a real design case.

Positional order: confirmed, but not fixed here. Thirteen merged formulas carry the same mismatch and one plain formula swaps its two arguments outright, so the fix belongs in `ComparisonFormula` rather than in these two classes. Measured and proposed in #1124.

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_70.py
```diff
@@ -0,0 +1,119 @@
...
+
+from blueprints.codes.eurocode.fpr_en_1992_1_1_2023 import FPR_EN_1992_1_1_2023
+from blueprints.codes.formula import ComparisonFormula
+from blueprints.codes.latex_formula import LatexFormula, latex_replace_symbols
+from blueprints.type_alias import DEG, DIMENSIONLESS, MPA
+from blueprints.utils.math_helpers import cot
+from blueprints.validations import raise_if_less_or_equal_to_zero, raise_if_negative
+
+
+class Form8Dot70CheckCrushingOfCompressionFieldInFlange(ComparisonFormula):
```
casper-van-der-krift: If this evaluates to True, that means that crushing is prevented. So perhaps update the class name to CheckPreventCrushingOfCompressionFieldInFlange.

casper-van-der-krift (changes requested):

blueprints/codes/eurocode/fpr_en_1992_1_1_2023/chapter_8_ultimate_limit_states/formula_8_70.py
egarciamendez: Formula evaluates if compressive stress is lower than compressive strength. So if that is True, the concrete is NOT crushing.

